PR #108, opened 2026-02-02
fix: correct YAML structure examples in skill documentation

stack72:
- Fix model input YAML examples in swamp-model skill to match actual implementation
- Fix workflow YAML example in swamp-workflow skill to match actual implementation
- Update job dependency syntax from deprecated `needs` to `dependsOn` with trigger conditions

## Details
The skill documentation examples were using a Kubernetes-style manifest structure with `apiVersion`, `kind`, and `metadata` fields
that don't exist in the actual implementation. The correct structure uses flat top-level fields (`id`, `name`, `version`, `tags`,
`attributes` for models; `id`, `name`, `description`, `version`, `jobs` for workflows).

## Test plan
- [ ] Review updated examples match actual YAML files created by `swamp model create` and `swamp workflow create`
- [ ] Verify examples are valid according to their respective schemas

### DIFF
--- a/.claude/skills/swamp-model/SKILL.md
+++ b/.claude/skills/swamp-model/SKILL.md
@@ -118,12 +118,11 @@ attributes according to the `inputAttributesSchema` from `type describe`.
 **Example input file structure:**
 
 ```yaml
-# inputs/swamp/echo/my-echo.yaml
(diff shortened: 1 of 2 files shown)


adamhjk:
this is hilarious

PR #211, opened 2026-02-08
fix: Unwrap single-artifact data in context builder

keeb:
## Summary

- When a model has exactly one data artifact, `model.X.data` now resolves
  directly to the `DataRecord` instead of wrapping it in a
  `{artifactName: DataRecord}` map. This fixes CEL expressions like
  `model.X.data.attributes.vpcId` that previously couldn't find
  `.attributes` on the outer map object.
- Multi-artifact models continue to use the map form for backward
  compatibility.
- Updated `ModelData.data` type to a union
  (`Record<string, DataRecord> | DataRecord`) and updated context building
  in both `model_resolver.ts` and `execution_service.ts`.

## Test Plan

- All 1602 existing tests pass with 0 failures
- Updated single-artifact tests to verify unwrapped `DataRecord` access
- Multi-artifact tests verified unchanged (still use map access)
- `deno check`, `deno lint`, `deno fmt` all pass
- `deno run compile` succeeds

### DIFF
--- a/integration/cel_data_access_test.ts
+++ b/integration/cel_data_access_test.ts
@@ -218,16 +218,19 @@ Deno.test("CEL Data Access: access latest data via model.X.data.Y", async () => 
     });
     const context = await modelResolver.buildContext();
 
-    // Access latest data
+    // Access latest data (single artifact unwrapped)
     const modelData = context.model["data_source"];
     assertExists(modelData);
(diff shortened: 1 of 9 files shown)


adamhjk:
Is this actually the right thing to do? Shouldn't the CEL expression be explicit?

PR #437, opened 2026-02-23
fix: redact vault secrets from persisted log files

### LINKED ISSUE #429: bug: Vault secrets are persisted in plaintext in data files and workflow run logs
## Summary

Vault secrets resolved via `vault.get()` CEL expressions are persisted in plaintext in two locations:

1. **Data files** — the `command` field in `.swamp/data/` contains the resolved secret value (e.g., `echo super-secret-value-12345` instead of the CEL template)
2. **Workflow run logs** — the resolved secret appears in plaintext in `.swamp/workflow-runs/<workflow-id>/workflow-run-<run-id>.log`

### Data file leakage

1. Create a vault and store a secret:
   ```
   swamp vault create local_encryption test-vault
   swamp vault put test-vault SECRET_VALUE=super-secret-value-12345
   ```
2. Install a model that references the secret via CEL:
   ```yaml
   # vault-echo model
   globalArguments:
     run: "echo ${{ vault.get('test-vault', 'SECRET_VALUE') }}"
   ```
3. Execute the model:
   ```
   swamp model method run vault-echo execute
   ```
4. Inspect persisted data:
   ```
   swamp data get vault-echo result --json
   ```
5. The `content.command` field contains `echo super-secret-value-12345` — the raw secret.

### Workflow log leakage

1. Same vault setup as above
2. Create a workflow that references the `vault-echo` model
3. Run the workflow:
   ```
   swamp workflow run vault-workflow
   ```
4. The resolved secret appears in plaintext in `.swamp/workflow-runs/<id>/workflow-run-<id>.log`

## Expected behavior

- The `command` field in persisted data should contain the CEL template (`echo ${{ vault.get('test-vault', 'SECRET_VALUE') }}`), not the resolved value
- Workflow run logs should redact or mask resolved vault secrets

## Impact

Any user or process with read access to the `.swamp/` directory can extract vault secrets in plaintext, defeating the purpose of vault encryption.

## References

- Related: #413, #418

stack72:
## Summary

- Vault secrets resolved via `vault.get()` were leaking into log files in plaintext (`.swamp/outputs/` and `.swamp/workflow-runs/`), defeating the purpose of vault encryption
- Adds a `SecretRedactor` that collects resolved vault secret values and replaces them with `***` in log output before writing to disk
- Redaction is applied only to **log files**, not data files — data file integrity is preserved for inter-step workflow data flow

### Why log-only redaction?

The original plan called for redacting both logs and data files. During implementation we identified that data file redaction would:
- **Break inter-step workflows** — downstream steps reading resource attributes via CEL would get `***` instead of actual values
- **Silently corrupt data** — models write data expecting it to be persisted as-is
- **Cause false positives** — common strings matching a secret value would be replaced in unrelated fields

Log files are audit/debug output where secrets should never appear. Data files are the model's responsibility. This narrower scope is the correct fix.

Closes #429

## Test plan

- New unit tests for `SecretRedactor` (8 tests covering basic redaction, short secret bypass, multiple secrets, longest-first ordering, JSON-escaped variants, empty state)
- All 1868 existing tests pass
- UAT test cases filed at systeminit/swamp-uat#26 for end-to-end verification

### DIFF
--- a/src/cli/commands/model_method_run.ts
+++ b/src/cli/commands/model_method_run.ts
@@ -38,6 +38,7 @@ import {
 import { coerceInputTypes, parseInputs } from "../input_parser.ts";
 import { parseTags } from "./data_search.ts";
 import { InputValidationService } from "../../domain/inputs/mod.ts";
+import { SecretRedactor } from "../../domain/secrets/mod.ts";
 import { join } from "@std/path";
 import {
   SWAMP_SUBDIRS,
@@ -152,6 +153,9 @@ export const modelMethodRunCommand = new Command()
       // Create run logger for real-time output
       const runLogger = getRunLogger(definition.name, methodName);
 
+      // Create secret redactor — populated during vault resolution, used by log sink and data writers
+      const redactor = new SecretRedactor();
+
       // Register run file sink target for log persistence
       const timestamp = new Date().toISOString().replace(/[:.]/g, "-");
       const logFilePath = join(
@@ -161,7 +165,7 @@ export const modelMethodRunCommand = new Command()
         `${definition.id}-${timestamp}.log`,
       );
       const runLogCategory: string[] = [];
-      await runFileSink.register(runLogCategory, logFilePath);
+      await runFileSink.register(runLogCategory, logFilePath, redactor);
 
       runLogger.info("Found model {name} ({type})", {
         name: definition.name,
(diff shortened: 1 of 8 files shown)


stack72:
## Follow-up: data file redaction

This PR intentionally scopes redaction to log files only. The data file problem (secrets persisting in `.swamp/data/`) is still real but requires a different approach.

Blanket string-replace redaction on data files causes silent data corruption, breaks inter-step workflow data flow, and produces false positives. The right solution is field-level redaction using the `sensitive: true` Zod metadata already described in `design/vaults.md`. This lets models explicitly declare which fields contain secrets, so:

- Only marked fields are redacted/omitted on disk
- In-memory values remain intact for inter-step data flow
- No false positives from common strings matching secret values
- Display layer (`swamp data get`) can mask sensitive fields

This is a larger piece of work touching the schema system, data writer, and display layer — it should be tracked as a separate issue.

PR #544, opened 2026-02-28
feat: add allowFailure flag on workflow steps

### LINKED ISSUE #541: Add allowFailure flag on workflow steps
## Problem

When building test or diagnostic workflows, some steps may fail due to external constraints (e.g., billing plan limitations, optional features not configured) rather than actual errors. Currently, any failed step marks the entire job and workflow as "failed", even when the failure is expected and acceptable.

There is no way to mark a step as "optional" or "allowed to fail." The existing dependency condition system (`completed`, `always`, etc.) controls whether downstream steps run, but doesn't prevent the failed step from counting toward the overall workflow status.

## Proposed Solution

Add an `allowFailure` (or `continueOnError`) boolean flag to the step schema:

```yaml
steps:
  - name: check-log-streaming
    description: Check log streaming config (may 403 on free plans)
    allowFailure: true
    task:
      type: model_method
      modelIdOrName: my-log-config
      methodName: get
      inputs:
        logType: configuration
```

When `allowFailure: true`:
- If the step succeeds, it reports as `succeeded` as normal
- If the step fails, it reports as something like `failed_allowed` or `warning` instead of `failed`
- The step's failure does NOT propagate to the job or workflow status
- Downstream steps with `dependsOn: succeeded` still skip (the step did fail), but `dependsOn: completed` would fire

## Use Case

We have a monolith test workflow (`test-all-models`) that exercises all 10 Tailscale model types in parallel. Two of the steps call `logConfig.get`, which returns HTTP 403 ("feature not available on current billing plan") on free-tier tailnets. The API call and error handling work correctly — it's just that the feature isn't available.

Without `allowFailure`, the workflow reports as "failed" even though 14 of 16 steps pass and the 2 failures are expected. This makes the workflow unusable as an automated health check because the exit code is always non-zero.

## Alternatives Considered

- **Remove the steps**: Works but loses visibility into which features are available
- **Restructure into separate workflows**: Adds complexity without solving the underlying problem
- **Use dependency conditions**: `completed`/`always` let downstream steps run, but the overall workflow status is still "failed"

stack72:
## Summary

Closes #541

- Adds `allowFailure: boolean` (default `false`) to the step schema, allowing steps to fail without propagating failure to the job or workflow
- When a step with `allowFailure: true` fails, it is recorded as failed with `allowedFailure: true` on the StepRun, but does not cause the job to fail
- Trigger condition behavior is unchanged: `succeeded` → false, `failed` → true, `completed` → true — so downstream steps using `dependsOn: completed` still fire
- Output rendering shows a `⚠` icon for allowed-failure steps instead of the regular `✗`

## Design

The implementation leverages the existing `Promise.allSettled` pattern in the execution loop. When a step with `allowFailure: true` catches an error, it marks the StepRun as an allowed failure and returns normally (instead of rethrowing). This means `Promise.allSettled` sees a fulfilled promise, and `jobFailed` is never set — so no changes were needed to the job execution loop itself.

## Changes

- `src/domain/workflows/step.ts` — Added `allowFailure` to schema, class, and serialization
- `src/domain/workflows/workflow_run.ts` — Added `allowedFailure` tracking on `StepRun`
- `src/domain/workflows/execution_service.ts` — Modified catch blocks in both `executeStep()` and `executeExpandedStep()` to handle `allowFailure`
- `src/domain/workflows/job.ts` / `workflow.ts` — Added `Input` types for backward-compatible `fromData()`
- `src/presentation/output/workflow_run_output.ts` — Added `⚠` icon for allowed-failure steps
- `src/cli/commands/workflow_run.ts`, `workflow_history_get.ts`, `workflow_run_search.ts`, `workflow_history_search.ts` — Thread `allowedFailure` through to output
- `design/workflow.md` — Documented behavior
- `.claude/skills/swamp-workflow/SKILL.md` — Added usage example

## Test plan

- [x] Unit tests for `Step` — `allowFailure` defaults to false, round-trips through create/fromData/toData
- [x] Unit tests for `StepRun` — `markAllowedFailure()` sets flag, serializes only when true
- [x] Execution service tests:
  - [x] Step with `allowFailure: true` fails → job still succeeds
  - [x] Step with `allowFailure: true` fails → workflow still succeeds
  - [x] Downstream step with `dependsOn: succeeded` skips when `allowFailure` step fails
  - [x] Downstream step with `dependsOn: completed` runs when `allowFailure` step fails
  - [x] Mix of `allowFailure` and regular failing steps → job fails (regular failure dominates)
- [x] All 2366 tests pass
- [x] `deno check`, `deno lint`, `deno fmt` pass
- [x] Binary compiled and manually tested end-to-end (see PR comment for example workflow)

### DIFF
--- a/src/domain/workflows/execution_service.ts
+++ b/src/domain/workflows/execution_service.ts
@@ -17,7 +17,7 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with Swamp.  If not, see <https://www.gnu.org/licenses/>.
 
-import { Workflow, type WorkflowData, type WorkflowInput } from "./workflow.ts";
+import { Workflow, type WorkflowInput } from "./workflow.ts";
 import type { Job } from "./job.ts";
 import type { Step } from "./step.ts";
 // deno-lint-ignore verbatim-module-syntax
--- a/src/domain/workflows/job.ts
+++ b/src/domain/workflows/job.ts
@@ -23,7 +23,7 @@ import {
   type TriggerConditionData,
   TriggerConditionSchema,
 } from "./trigger_condition.ts";
-import { Step, type StepData, type StepInput, StepSchema } from "./step.ts";
+import { Step, type StepData, StepSchema } from "./step.ts";
 
 /**
  * Schema for job dependency with condition.
--- a/src/domain/workflows/workflow.ts
+++ b/src/domain/workflows/workflow.ts
@@ -19,7 +19,7 @@
 
 import { z } from "zod";
 import { createWorkflowId, type WorkflowId } from "./workflow_id.ts";
-import { Job, type JobData, type JobInput, JobSchema } from "./job.ts";
+import { Job, type JobData, JobSchema } from "./job.ts";
 import {
   type InputsSchema,
   InputsSchemaSchema,


stack72:
## End-to-end test with compiled binary

I created a test repo in `/tmp` and ran this workflow with the compiled binary to verify the feature works as described in the issue.

### Example workflow file

```yaml
id: fae3577e-658c-4e56-a146-e44321ab4810
name: allow-failure-test
description: Test workflow demonstrating allowFailure behavior
tags: {}
jobs:
  - name: main
    description: Job with an allowed-to-fail step
    steps:
      - name: optional-check
        description: This step will fail but is allowed to
        allowFailure: true
        task:
          type: model_method
          modelIdOrName: nonexistent-model
          methodName: execute
        dependsOn: []
        weight: 0
      - name: always-runs
        description: This step runs regardless of optional-check result
        task:
          type: model_method
          modelIdOrName: will-succeed
          methodName: execute
          inputs:
            run: "echo hello world"
        dependsOn:
          - step: optional-check
            condition:
              type: completed
        weight: 0
    dependsOn: []
    weight: 0
version: 1
```

### Result

The `optional-check` step fails (model doesn't exist) but is marked as an **allowed failure**. The `always-runs` step executes successfully because it depends on `optional-check` with `condition: completed` (not `succeeded`). The overall workflow reports as **succeeded**.

Output shows the `⚠` icon for allowed-failure steps instead of the regular `✗`:

```
⚠ optional-check  (failed — allowed failure)
✓ always-runs     (succeeded)
```

This matches the use case from the issue — steps that may fail due to external constraints (e.g., billing plan limits) can be marked with `allowFailure: true` so they don't cause the entire workflow to fail.

PR #584, opened 2026-03-03
fix: correct Kiro hook format and add USER_PROMPT env var support

### LINKED ISSUE #581: Swamp hooks not formatted properly for Kiro, Kiro doesn't pipe and details into it's PostToolUse hook
I saw that the new hook is created for kiro when I used `swamp repo upgrade --tool kiro` and noticed that the hook was not present in the UI. After mucking around with kiro a bit, it looks like there are two issues:

1. Kiro wants a "toolTypes" property (an array of strings, supporting wild cards, so the base one we could start with is `"toolTypes": ["*"]` as part of the `when` property
2. Kiro wants the extension to be `swamp-audit.kiro.hook`, not `.json`

Once those changes were made, the hook appeared as expected. Kiro makes the prompt available in the `USER_PROMPT` env variable, it does _not_ pipe it into the hook via stdin (debugged and got it to work by having kiro write a wrapping script that piped it in)

stack72:
## Summary

Fixes #581 — Kiro IDE does not recognize the swamp audit hook due to three issues, all addressed in this PR.

### 1. Hook file format
- **Wrong extension**: Kiro expects `.kiro.hook`, not `.json` — renamed `swamp-audit.json` → `swamp-audit.kiro.hook`
- **Missing `toolTypes`**: Kiro requires `toolTypes: ["*"]` in the `when` property — added
- **No timeout**: Added `timeout: 5` to the `then` property to prevent hangs as a safety net

### 2. Input mechanism — `USER_PROMPT` env var
Kiro IDE passes `postToolUse` data via the `USER_PROMPT` environment variable, **not** stdin. The `audit record` command was blocking on `readStdin()` which caused timeouts. The fix:
- For `--tool kiro`: check `USER_PROMPT` env var first, fall back to stdin (preserving kiro-cli compatibility)
- The Kiro IDE format uses camelCase keys (`toolName`, `toolArgs`, `toolResult`, `toolSuccess`) which differs from the kiro-cli snake_case format

### 3. Dual-format normalizer
Updated `normalizeKiro()` to handle both:
- **kiro-cli** (stdin, snake_case): `tool_name`, `tool_input`, `tool_response`
- **Kiro IDE** (`USER_PROMPT`, camelCase): `toolName`, `toolArgs`, `toolResult`, `toolSuccess`

### 4. Upgrade cleanup
`swamp repo upgrade --tool kiro` now removes the old `swamp-audit.json` file so stale hooks don't linger.

## Open question for @danmcclain

The `USER_PROMPT` env var format is based on Kiro's own description of its `postToolUse` data. Two things we'd appreciate verification on:

1. **Is `USER_PROMPT` valid JSON?** We parse it with `JSON.parse()` — if it's a different format, the hook will silently no-op (safe but won't record).
2. **Does `toolArgs` contain `{ command: "..." }` for `execute_bash` tools?** You noted it "appears to be empty `{}`" in your examples — if that's always the case, we won't be able to extract the command. Could you verify with your test hook on an `execute_bash` invocation?

## Files changed

| File | Change |
|------|--------|
| `src/domain/repo/repo_service.ts` | `.kiro.hook` extension, `toolTypes`, `timeout`, old file cleanup |
| `src/cli/commands/audit.ts` | `readHookInput()` with `USER_PROMPT` env var fallback |
| `src/domain/audit/hook_input.ts` | Dual-format `normalizeKiro()` |
| `src/domain/audit/hook_input_test.ts` | 4 new tests for Kiro IDE camelCase format |
| `src/domain/repo/repo_service_test.ts` | Updated hook format tests + old file cleanup test |

## Test plan

- [x] `deno check` passes
- [x] `deno lint` passes
- [x] `deno fmt` passes
- [x] `deno run test` — 2605 tests pass (4 new)
- [x] `deno run compile` — binary builds
- [x] Manual test: `swamp repo init --tool kiro` creates `swamp-audit.kiro.hook` with correct format
- [ ] @danmcclain to verify with Kiro IDE that the hook appears and `USER_PROMPT` format works

### DIFF
--- a/src/cli/commands/audit.ts
+++ b/src/cli/commands/audit.ts
@@ -49,6 +49,21 @@ async function readStdin(): Promise<string> {
   return chunks.join("");
 }
 
+/**
+ * Reads hook input for the given tool.
+ *
+ * Kiro IDE passes postToolUse data via the USER_PROMPT environment variable
+ * rather than stdin. We check USER_PROMPT first for kiro, falling back to
+ * stdin for kiro-cli compatibility. All other tools read from stdin.
+ */
+function readHookInput(tool: HookTool): Promise<string> {
+  if (tool === "kiro") {
+    const envInput = Deno.env.get("USER_PROMPT");
+    if (envInput) return Promise.resolve(envInput);
+  }
+  return readStdin();
+}
+
 /** Valid values for the --tool option */
 const VALID_HOOK_TOOLS: HookTool[] = ["claude", "cursor", "kiro", "opencode"];
 
@@ -76,7 +91,7 @@ export const auditRecordCommand = new Command()
         return;
       }
 
-      const input = await readStdin();
+      const input = await readHookInput(tool);
       if (!input.trim()) {
         return;
       }
--- a/src/domain/audit/hook_input.ts
+++ b/src/domain/audit/hook_input.ts
@@ -141,23 +141,42 @@ function normalizeCursor(
 
 /**
  * Kiro: postToolUse hook (fires for both success and failure).
- * tool_name === "execute_bash", command from tool_input.command,
- * failure when tool_response.success === false.
+ *
+ * Supports two input formats:
+ * - kiro-cli (stdin, snake_case): tool_name, tool_input, tool_response
+ * - Kiro IDE (USER_PROMPT env var, camelCase): toolName, toolArgs, toolResult, toolSuccess
+ *
+ * tool_name/toolName === "execute_bash", command from tool_input/toolArgs,
+ * failure when tool_response.success === false or toolSuccess === false.
  */
 function normalizeKiro(
   raw: Record<string, unknown>,
 ): NormalizedHookInput | null {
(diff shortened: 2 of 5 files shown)


stack72:
There's a follow up issue here due to Kiro IDE:

```
Yes, this is a bug in Kiro IDE (the desktop app).

The Kiro IDE is not populating the toolArgs field in the USER_PROMPT environment variable when running postToolUse hooks. It's sending:

{"toolName":"executeBash","toolArgs":{},"toolResult":"...","toolSuccess":true}
But it should be sending:

{"toolName":"executeBash","toolArgs":{"command":"date"},"toolResult":"...","toolSuccess":true}
Since you confirmed that kiro-cli works correctly with the same hook configuration, this is specifically a bug in the desktop IDE's hook implementation. The hook system needs to include the actual tool parameters that were used when invoking the tool, not just an empty object.

Would you like me to help you report this bug to the Kiro team, or do you have a process for that already?
```

This means that we can;'t get audit support in Kiro IDE only Kiro CLI

PR #640, opened 2026-03-06
fix: track resource deletion in local data model (#637)

### LINKED ISSUE #637: Track resource deletion in local data model
## Problem

Swamp's local data store has no concept of resource deletion. When a delete method succeeds, the data from prior create/get/update calls persists with no marker indicating the resource is gone. This causes two related problems:

### 1. Operations on deleted resources make doomed API calls

After deleting a project, running `update` on the same model sends a PATCH to the DO API using the ID from local data, and gets a raw 404 back. Swamp has the knowledge that a delete succeeded (it just ran one), but doesn't use it. The user sees a confusing API error instead of a clear "resource was deleted" message.

### 2. Re-create after delete mixes version history

After deleting a volume and re-creating it, the version counter continues from where it left off:

- v1: `id: 8290fa55` (deleted volume)
- v2: `id: 8290fa55` (deleted volume, refreshed via get)  
- v3: `id: 407a6757` (new volume, completely different cloud resource)

`data.latest()` works fine, but `data.version("model", "instance", 1)` returns data for a resource that no longer exists. There's no boundary between the old and new resource. Anyone iterating versions for audit or rollback would be mixing two unrelated cloud resources.

## Proposed solution

Record a lifecycle event when a delete method succeeds. This could be:

- A deletion marker in the data store (e.g. a version entry with `"lifecycle": "deleted"`)
- A metadata flag on the instance (e.g. `"deletedAt": "2026-03-06T01:02:00Z"`)

This would enable:
- Fast-fail on update/get with a clear message: "resource was deleted at time T, run create first"
- Version history boundaries: versions before the marker belong to a previous resource
- Optional version counter reset on re-create after delete

## Reproduction

```bash
# Create, then delete
swamp model method run test-project create --json
swamp model method run test-project delete --input '{"id": "<project-id>"}' --json

# Update gives raw 404 instead of "resource deleted" error
swamp model method run test-project update --json

# Re-create continues version counter, mixing old and new resource data
swamp model method run test-volume create --json  # stores as v3, not v1
```

## Scope

This is a swamp core issue affecting the data store and method execution path. Changes would be needed in:
- Data writer: record deletion events
- Method execution: check for deletion markers before making API calls
- Version history: insert lifecycle boundaries on delete

stack72:
## Problem

When a user runs a `delete` method on a model (e.g., `swamp model method run my-widget delete`), the cloud resource is destroyed but swamp's local data store retains no record of the deletion. This creates two problems:

1. **Doomed API calls**: Subsequent `get` or `update` methods happily execute against a resource that no longer exists, resulting in raw cloud provider errors (HTTP 404, "resource not found", etc.) that are confusing and unhelpful to the user.

2. **No deletion history**: There is no way to distinguish "this resource was never created" from "this resource existed but was deleted" — the data store simply shows the last active version with no indication that the resource is gone.

### User Impact

Before this fix, a typical user experience looks like:

```
$ swamp model method run my-widget create    # ✅ Creates the resource
$ swamp model method run my-widget delete    # ✅ Deletes the resource
$ swamp model method run my-widget get       # ❌ Raw 404 error from cloud API
$ swamp model method run my-widget update    # ❌ Raw 404 error from cloud API
```

After this fix:

```
$ swamp model method run my-widget create    # ✅ Creates the resource
$ swamp model method run my-widget delete    # ✅ Deletes + writes deletion marker
$ swamp model method run my-widget get       # ❌ Clear error: "Resource 'widget' was deleted at 2026-03-06T19:00:00Z — run a 'create' method to re-create it first"
$ swamp model method run my-widget update    # ❌ Same clear error
$ swamp model method run my-widget create    # ✅ Re-creates, clears deletion state
$ swamp model method run my-widget get       # ✅ Works again
```

### 1. Method Kind Classification (`MethodKind` + `inferMethodKind()`)

Added a `MethodKind` type (`"create" | "read" | "update" | "delete" | "list" | "action"`) and an `inferMethodKind()` utility that:
- Returns the explicit `kind` if set on the method definition
- Otherwise infers from conventional names: `create`→create, `get/read/describe/show`→read, `update/patch`→update, `delete/destroy/remove`→delete, `list/search/find`→list
- Returns `undefined` for unrecognized names

Extension models can also set `kind` explicitly on their method definitions for non-conventional names.

### 2. Data Lifecycle Tracking (`DataLifecycle`)

Added a `lifecycle` field to `DataMetadata` with values `"active"` (default) or `"deleted"`. This is backward-compatible — existing data without a lifecycle field is treated as active.

### 3. Deletion Markers (Tombstones)

After a `delete` method succeeds, `executeWorkflow()` writes a new version of each resource with `lifecycle: "deleted"` and JSON content containing `{ deletedAt, deletedByMethod }`. This acts as a tombstone that records when and how the resource was deleted.

### 4. Fast-Fail on Deleted Resources

Before executing `read` or `update` methods, `executeWorkflow()` checks for deletion markers. If any resource has `lifecycle: "deleted"`, it throws a `UserError` with a clear message including the deletion timestamp and instructions to re-create.

### 5. Create Clears Deletion State

When a `create` method writes new resource data, `Data.create()` defaults `lifecycle` to `"active"`, naturally superseding the deletion marker. The `latest` symlink points to the new active version.

### 6. GC Interaction

Deletion markers (tombstones) are permanent — `findExpiredData()` in the data lifecycle service skips data with `lifecycle: "deleted"` so they are never auto-expired.

## Files Changed

| File | Change |
|------|--------|
| `src/domain/models/model.ts` | `MethodKind` type, `kind` field on `MethodDefinition`, `inferMethodKind()` utility |
| `src/domain/models/user_model_loader.ts` | Propagate `kind` through extension model loader |
| `src/domain/data/data_metadata.ts` | `DataLifecycleSchema`, `lifecycle` field in metadata |
| `src/domain/data/data.ts` | `lifecycle` in Data class, `isDeleted` accessor, `withDeletionMarker()` factory |
| `src/domain/data/mod.ts` | Export new types |
| `src/domain/models/method_execution_service.ts` | Pre-check for deleted resources, post-delete marker writing |
| `src/domain/data/data_lifecycle_service.ts` | Skip deletion markers in GC |

### Unit Tests (33 new tests)

**`model_test.ts`** (8 tests)
- `inferMethodKind` returns explicit kind from definition
- Infers `create`, `read`, `update`, `delete`, `list` from conventional names
- Returns `undefined` for unrecognized method names
- Explicit `kind` overrides conventional name inference

**`data_metadata_test.ts`** (4 tests)
- Schema accepts `"active"` and `"deleted"` lifecycle values
- Schema rejects invalid lifecycle values
- Schema accepts `undefined` (backward compatibility)

**`data_test.ts`** (12 tests)
- Defaults lifecycle to `"active"` on create
- Creates data with `"deleted"` lifecycle
- `isDeleted` returns true/false correctly
- `withDeletionMarker()` creates correct deletion marker
- `toData()` includes lifecycle only when `"deleted"` (saves disk)
- `fromData()` defaults to `"active"` when lifecycle missing
- Roundtrip preserves `"deleted"` lifecycle
- `withNewVersion()` preserves lifecycle

**`method_execution_service_test.ts`** (6 tests)
- Delete method writes deletion markers to all resources
- Read after delete throws `UserError` with timestamp
- Update after delete throws `UserError`
- Create after delete succeeds (clears deletion state)
- Delete skips already-deleted resources
- Explicit `kind` override prevents deletion marker writing

**`data_lifecycle_service_test.ts`** (2 tests)
- Skips deletion markers in `findExpiredData()`
- Returns expired active data alongside deletion markers

**`user_model_loader_test.ts`** (1 test)
- `kind` propagated through model definition conversion

### E2E Verification

Manually verified the full lifecycle with a test extension model:
1. Create → version 1 (active) ✅
2. Get → succeeds ✅
3. Delete → writes version 2 with `lifecycle: deleted` ✅
4. Get after delete → fails with `UserError` ✅
5. Update after delete → fails with `UserError` ✅
6. Create after delete → writes version 3 (active) ✅
7. Get after re-create → succeeds ✅

## Test plan

- [x] `deno check` — Type checking passes
- [x] `deno lint` — No lint errors
- [x] `deno fmt` — Formatted
- [x] `deno run test` — All 2765 tests pass
- [x] `deno run compile` — Binary compiles
- [x] E2E verification of full create → delete → fail → re-create lifecycle

Closes #637

### DIFF
--- a/src/domain/data/data.ts
+++ b/src/domain/data/data.ts
@@ -19,6 +19,7 @@
 
 import { createDataId, type DataId, generateDataId } from "./data_id.ts";
 import {
+  type DataLifecycle,
   type DataMetadata,
   DataMetadataSchema,
   type GarbageCollectionPolicy,
@@ -43,6 +44,7 @@ export interface CreateDataProps {
   createdAt?: Date;
   size?: number;
   checksum?: string;
+  lifecycle?: DataLifecycle;
 }
 
 /**
@@ -71,6 +73,7 @@ export class Data {
     readonly tags: Record<string, string>,
     readonly ownerDefinition: OwnerDefinition,
     readonly createdAt: Date,
+    readonly lifecycle: DataLifecycle,
     readonly size?: number,
     readonly checksum?: string,
   ) {}
@@ -87,6 +90,7 @@ export class Data {
     const version = props.version ?? 1;
     const createdAt = props.createdAt ?? new Date();
     const lifetime = normalizeLifetime(props.lifetime);
+    const lifecycle = props.lifecycle ?? "active";
 
     const validated = DataMetadataSchema.parse({
       id,
@@ -101,6 +105,7 @@ export class Data {
       createdAt: createdAt.toISOString(),
       size: props.size,
       checksum: props.checksum,
+      lifecycle: lifecycle === "active" ? undefined : lifecycle,
     });
 
     return new Data(
@@ -114,6 +119,7 @@ export class Data {
       { ...validated.tags },
       { ...validated.ownerDefinition },
       new Date(validated.createdAt),
+      validated.lifecycle ?? "active",
       validated.size,
       validated.checksum,
     );
@@ -142,6 +148,7 @@ export class Data {
       { ...validated.tags },
       { ...validated.ownerDefinition },
       new Date(validated.createdAt),
+      validated.lifecycle ?? "active",
       validated.size,
       validated.checksum,
     );
@@ -170,6 +177,9 @@ export class Data {
     if (this.checksum !== undefined) {
       data.checksum = this.checksum;
     }
+    if (this.lifecycle === "deleted") {
+      data.lifecycle = "deleted";
+    }
 
     return data;
   }
@@ -191,6 +201,32 @@ export class Data {
    */
   get type(): string {
     return this.tags.type;
+  }
+
+  /**
+   * Returns true if this data has been marked as deleted.
+   */
+  get isDeleted(): boolean {
+    return this.lifecycle === "deleted";
+  }
+
+  /**
+   * Creates a deletion marker version of this data.
+   * The marker has lifecycle "deleted", JSON content type, and streaming disabled.
+   */
+  withDeletionMarker(props: { version: number }): Data {
+    return Data.create({
+      id: this.id,
+      name: this.name,
+      version: props.version,
+      contentType: "application/json",
+      lifetime: this.lifetime,
+      garbageCollection: this.garbageCollection,
+      streaming: false,
+      tags: this.tags,
+      ownerDefinition: this.ownerDefinition,
+      lifecycle: "deleted",
+    });
   }
 
   /**
@@ -216,6 +252,7 @@ export class Data {
       createdAt: props.createdAt ?? new Date(),
       size: props.size,
       checksum: props.checksum,
+      lifecycle: this.lifecycle,
     });
   }
 }
--- a/src/domain/data/data_lifecycle_service.ts
+++ b/src/domain/data/data_lifecycle_service.ts
@@ -177,6 +177,10 @@ export class DefaultDataLifecycleService implements DataLifecycleService {
 
     const allData = await this.dataRepo.findAllGlobal();
     for (const { data, modelType, modelId } of allData) {
+      // Skip deletion markers — they are tombstones and should not be auto-expired
+      if (data.isDeleted) {
+        continue;
+      }
       try {
         const isExpired = await this.isExpired(data);
         if (isExpired) {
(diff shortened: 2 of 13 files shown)


stack72:
## Why this is a swamp bug (not an extension model bug)

When a `delete` method succeeds, the cloud resource is destroyed — but swamp's local data store still shows the last active version with no record that deletion happened.

If a user (or a workflow) subsequently runs `get` or `update`, swamp happily executes the method against a resource that no longer exists. The cloud provider returns a raw error (404, "not found", etc.) that bubbles up unfiltered — confusing and unhelpful.

This is a swamp framework bug because:

- **Swamp owns the data layer** — it's swamp's responsibility to track the lifecycle state of resources it manages
- **Swamp owns method orchestration** — `executeWorkflow()` knows which method ran and whether it succeeded, so it's the right place to record that a deletion happened
- **Extension models shouldn't have to implement this** — every model would need to independently handle "don't call the API if the resource was deleted," which is boilerplate that belongs in the framework

The fix makes swamp's data layer lifecycle-aware: deletion markers (tombstones) are written after successful deletes, and read/update methods fast-fail with a clear message telling the user to re-create first.

PR #671, opened 2026-03-10
fix: increase extension max file count to 150

stack72:
## Summary

- The extension safety analyzer's `MAX_FILE_COUNT` was set to 100, which blocked pushing larger service extensions like `@swamp/aws/ec2` (104 models)
- Increased the limit from 100 to 150 to accommodate extensions that cover broad service APIs
- Updated the corresponding test, design doc, and skill reference to match

## Test Plan

- [x] Updated test now creates 151 files and asserts the limit is 150
- [x] All 111 extension domain tests pass
- [x] `deno lint` passes

### DIFF
--- a/.claude/skills/swamp-extension-model/references/publishing.md
+++ b/.claude/skills/swamp-extension-model/references/publishing.md
@@ -190,7 +190,7 @@ The safety analyzer scans all files before push. Issues are classified as
 | Disallowed extensions       | Only `.ts`, `.json`, `.md`, `.yaml`, `.yml`, `.txt` |
 | File too large              | Individual files must be under 1 MB                 |
 | Total size exceeded         | All files combined must be under 10 MB              |
-| Too many files              | Maximum 100 files per extension                     |
+| Too many files              | Maximum 150 files per extension                     |
 
 ### Warnings (prompted)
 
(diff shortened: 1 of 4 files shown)


stack72:
Hey @umag — thanks for the detailed investigation and POC! The A/B benchmarks and root-cause analysis made this very easy to triage.

## What we explored

We looked at three approaches:

1. **Detached subprocess flush** — spawn a background child process to handle the HTTP POST while the parent exits immediately. This looked promising, but the compiled binary's cold-start (~200-300ms) means the child can't finish flushing within any timeout shorter than its own startup time. Effectively fire-and-forget, which loses telemetry in ephemeral repos (CI, containers) where there's no next invocation.

2. **Subprocess + short grace window** — spawn the child but await it for ~200ms before detaching. Same cold-start problem: the child hasn't even begun flushing by the time the window closes.

3. **Tighter inline timeout** (what we went with) — reduce the existing `AbortSignal.timeout` from 2s to 200ms. Same code path, same spool-and-retry safety net, just a tuning change. Fast/local endpoints flush within the window; slow endpoints fall back to flushing on the next invocation, which the code already handles.

## What's still on the table

Your POC's approach (detached child process) could still be valuable if paired with a **lightweight flusher** that avoids the full binary cold-start — e.g. a small standalone script that only loads the sender and spool, not the entire CLI. That would get the flush truly off the critical path while preserving delivery. Worth revisiting if the 200ms window proves too tight for some endpoint configurations.

PR: https://github.com/swamp-club/swamp/pull/1594

PR #697, opened 2026-03-12
feat: add pre-flight checks for model method execution

### LINKED ISSUE #619: Support pre-flight validate method on models before method execution
## Problem

When a model method like `create` is executed, it goes straight to the API call with no pre-flight checks. If the inputs are invalid (e.g. a VPC CIDR range that overlaps with existing VPCs, or a region that doesn't support a given droplet size), the user only finds out via a raw API error response that is difficult to interpret.

**Real-world scenario encountered:**
1. User configures a VPC with CIDR `10.116.0.0/24`
2. Runs `create` method
3. DigitalOcean API returns a 422 with a JSON blob about overlapping ranges
4. User has to parse the raw error to understand the problem
5. This could have been caught before the API call by listing existing VPCs and checking for CIDR overlap

## Proposed Solution

If a model defines a `validate` method, swamp should automatically call it before executing the requested method (e.g. `create`, `update`). This gives models a hook to run pre-flight checks against live API state.

**How it would work:**

- Model authors define an optional `validate` method on their model
- Before executing any other method, swamp checks if `validate` exists and runs it first
- If `validate` fails (throws an error or returns validation errors), the method execution is aborted with a clear error message
- If `validate` succeeds, the requested method proceeds as normal

**Example use cases for validate:**
- VPC: Check existing VPCs for CIDR overlap before creating
- Droplet: Verify the image slug and size are available in the target region
- Load Balancer: Verify the target tag exists and has matching droplets
- Firewall: Verify referenced load_balancer_uids exist

## Scope of Changes

- **Model execution engine** — Add a pre-execution phase that checks for and runs a `validate` method before the target method
- **Extension model contract** — Document the `validate` method convention and what it should return (pass/fail with messages)
- **Workflow engine** — Ensure validation errors are surfaced clearly in workflow step output, distinct from execution errors
- **Code generation** — Auto-generated models could include common validation patterns (e.g. "does this resource name already exist")

This would catch configuration errors early with clear messages, rather than relying on raw API error responses after the fact.

stack72:
## Summary

Adds pre-flight checks that run automatically before mutating model methods (create/update/delete/action), catching invalid inputs early with clear error messages instead of letting them fail as raw API errors.

Closes #619

### Core Implementation
- `CheckDefinition` interface with `description`, `labels`, `appliesTo`, and `execute(context) => CheckResult`
- `CheckResult` with `pass: boolean` and optional `errors: string[]`
- Checks run automatically before mutating methods; read/list methods skip checks
- `isMutatingKind()` helper for method kind classification

### Extension Checks
- `ModelRegistry.extend()` accepts optional third `checks` parameter
- Extension checks merge with model-type checks; conflicts throw at registration time
- Extension checks format matches `methods`: array of `Record<string, CheckDefinition>`

### Definition-Level Check Selection
- `CheckSelection` type: `{ require?: string[]; skip?: string[] }` on definitions
- `skip` always wins (even over `require`)
- `require` checks are immune to CLI skip flags (`--skip-checks`, `--skip-check`, `--skip-check-label`)
- Validation warns on nonexistent check names and require/skip overlap

### CLI Flags
- `model method run`: `--skip-checks`, `--skip-check <name>`, `--skip-check-label <label>`
- `model validate`: `--label <label>`, `--method <method>` for check filtering

### Validation Service
- Check results appear alongside schema validations in `model validate` output
- Label and method filtering for targeted check execution
- Definition-level skip lists honored during validation

## Test plan

- [x] 17 new unit tests covering all check behaviors (pass/fail, filtering, skip flags, extension merging, definition-level selection)
- [x] Integration tests updated for new validation count
- [x] All 2875 tests pass
- [x] `deno check`, `deno lint`, `deno fmt` pass
- [x] End-to-end tested with `command/shell` model extension checks:
  - Check pass/fail on method execution
  - Label-based filtering (`--skip-check-label`)
  - Named check skipping (`--skip-check`)
  - Skip all checks (`--skip-checks`)
  - Definition-level require (immune to skip flags)
  - Definition-level skip (overrides require)
  - `model validate` with `--label` and `--method` filtering
  - Invalid check selection validation warnings

### DIFF
--- a/src/domain/models/method_execution_service.ts
+++ b/src/domain/models/method_execution_service.ts
@@ -331,7 +331,14 @@ export class DefaultMethodExecutionService implements MethodExecutionService {
                 tags: currentDefinition.tags,
               },
             });
-            if (!checkResult.pass) {
+            if (!checkResult || typeof checkResult.pass !== "boolean") {
+              failures.push({
+                checkName,
+                errors: [
+                  "Check returned invalid result (expected { pass: boolean })",
+                ],
+              });
+            } else if (!checkResult.pass) {
               failures.push({
                 checkName,
                 errors: checkResult.errors ?? ["Check failed"],
--- a/src/domain/models/method_execution_service_test.ts
+++ b/src/domain/models/method_execution_service_test.ts
@@ -2360,3 +2360,24 @@ Deno.test("executeWorkflow - required check still respects appliesTo", async () 
     "create-only-required",
   );
 });
+
+Deno.test("executeWorkflow - check returning invalid result treated as failure", async () => {
+  const service = new DefaultMethodExecutionService();
+  const model = createCheckModel({
+    "bad-check": {
+      description: "Returns garbage",
+      execute: () => Promise.resolve(undefined as unknown as { pass: boolean }),
+    },
+  });
+
+  const definition = Definition.create({
+    name: "test-definition",
+    globalArguments: {},
+  });
+  const { context } = createTestContext({ modelType: model.type });
+  await assertRejects(
+    () => service.executeWorkflow(definition, model, "create", context),
+    UserError,
+    "invalid result",
+  );
+});
--- a/src/domain/models/validation_service.ts
+++ b/src/domain/models/validation_service.ts
@@ -288,12 +288,16 @@ export class DefaultModelValidationService implements ModelValidationService {
         continue;
       }
 
-      // Filter by method if provided
-      if (
-        checkContext.method && check.appliesTo &&
-        !check.appliesTo.includes(checkContext.method)
-      ) {
-        continue;
+      // Filter by method: if --method given, only run matching checks.
+      // If no --method given, skip checks that have an explicit appliesTo
+      // (they only make sense in the context of a specific method).
+      if (check.appliesTo) {
+        if (!checkContext.method) {
+          continue;
+        }
+        if (!check.appliesTo.includes(checkContext.method)) {
+          continue;
+        }
       }
 
       try {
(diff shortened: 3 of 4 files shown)


stack72:
This allows us to do this:

```
  export const extension = {
    type: "command/shell",
    methods: [],
    checks: [
      {
        "no-dangerous-commands": {
          description: "Ensures the shell command does not contain dangerous patterns",
          labels: ["offline", "safety"],
          appliesTo: ["execute"],
          async execute(context) {
            const dangerous = ["rm -rf /", "mkfs.", "dd if=", ":(){"];
            const errors = [];
            const globalArgs = context.globalArgs;

            for (const [key, value] of Object.entries(globalArgs)) {
              if (typeof value === "string") {
                for (const pattern of dangerous) {
                  if (value.includes(pattern)) {
                    errors.push(
                      `globalArguments.${key} contains dangerous pattern: "${pattern}"`,
                    );
                  }
                }
              }
            }
            return errors.length > 0 ? { pass: false, errors } : { pass: true };
          },
        },

        "command-not-empty": {
          description: "Ensures the execute method has a non-empty run argument",
          labels: ["offline", "validation"],
          async execute(context) {
            const defRepo = context.definitionRepository;
            const defInfo = context.definition;
            const def = await defRepo.findByName(context.modelType, defInfo.name);
            if (def) {
              const run = def.methodData?.execute?.arguments?.run;
              if (!run || (typeof run === "string" && run.trim() === "")) {
                return { pass: false, errors: ["No 'run' argument set for execute method"] };
              }
            }
            return { pass: true };
          },
        },

        "slow-api-check": {
          description: "Simulates a slow API check (for testing skip functionality)",
          labels: ["api", "slow"],
          appliesTo: ["execute"],
          async execute() {
            return { pass: true };
          },
        },
      },
    ],
  };
```



  CLI usage:

```
  # Checks run automatically before mutating methods
  swamp model method run my-model execute

  # Skip all checks
  swamp model method run my-model execute --skip-checks

  # Skip by name or label
  swamp model method run my-model execute --skip-check slow-api-check
  swamp model method run my-model execute --skip-check-label api

  # Validate with filtering
  swamp model validate my-model --label offline
  swamp model validate my-model --method execute
```

If you want to encode the checks in the definition:

```
  name: my-model
  type: command/shell
  checks:
    require:
      - no-dangerous-commands
    skip:
      - slow-api-check
```

PR #713, opened 2026-03-15
feat: per-model datastore locks for concurrent model operations

### LINKED ISSUE #706: Datastore lock: allow concurrent reads across different models
## Problem

Currently the datastore uses a single global file lock. Running `swamp model method run` on one model blocks all other model method runs, even on completely unrelated models. This makes it impossible to run operations on different models in parallel (e.g. querying prod-vms while also running a workflow on a different model).

## Context

When running fleet operations across 100+ VMs, each method call can take 30-90 seconds. Being unable to run a quick query on a different model while a long operation is in progress feels unnecessarily rigid.

## Proposed Solution

Consider a more granular locking strategy:
- **Per-model locks** — each model instance gets its own lock, so unrelated models don't block each other
- **Read/write locks** — read-only operations (like `get`, `list`) could run concurrently, with exclusive locks only for writes
- **Per-method locks** — even within a model, independent methods on different resources could potentially run concurrently

The simplest improvement would be per-model locking, so that running a method on model A doesn't block an unrelated method on model B.

## Current behavior

```
# Terminal 1: starts a 90-second healthCheck on prod-vms
swamp model method run prod-vms healthCheck --json

# Terminal 2: tries to check a different model — blocked until Terminal 1 finishes
swamp model method run dev-vms checkService --input service=docker --json
# ^ waits for datastore lock, eventually times out
```

## Expected behavior

Operations on different model instances should not block each other.

stack72:
## Summary

Replaces the single global datastore lock with per-model locks so that operations on different models can run concurrently. This fixes the core issue where running `model method run` on one model blocked **all** other model operations, even on completely unrelated models.

- Introduces per-model lock files at `data/{modelType}/{modelId}/.lock` — operations on different models no longer block each other
- Refactors the sync coordinator from module-level singletons to a `Map<string, SyncEntry>` supporting multiple concurrent locks
- Adds model-scoped S3 sync (`pullChangedForModel`) so S3 datastores only pull/push the relevant model's files
- Extracts model references from workflows to acquire only the locks needed, with automatic fallback to global lock for dynamic CEL references
- Updates breakglass commands (`lock status`, `lock release --model`) to show and manage per-model locks

Closes #706

## Why per-model locks?

The global lock was designed for simplicity, but it creates an O(n) bottleneck for fleet operations. When managing 100+ VMs where each method call takes 30-90 seconds, a single global lock means:

- **Before**: Running `healthCheck` on `prod-vm-1` blocks a quick `checkService` on `dev-vm-2` — the second command waits or times out
- **After**: Both run concurrently because they acquire separate locks (`data/command/shell/prod-vm-1/.lock` and `data/command/shell/dev-vm-2/.lock`)

Structural commands (`data gc`, `repo init`, model create/delete) still use the global lock, and per-model lock acquisition checks for a held global lock first — so data integrity is preserved.

## Impact on users

- **`model method run`** on different models runs concurrently (the primary win)
- **`workflow run`** acquires only the locks for models referenced in the workflow
- **`model evaluate`** and **`workflow evaluate`** use per-model locks for single-model/workflow evaluation
- **S3 datastores** benefit equally — model-scoped pull on lock acquisition, brief global lock only during push (seconds, not minutes)
- **No breaking changes** — the lock file format and CLI interface are unchanged; existing repos work without migration
- **Deadlock prevention** — locks are always acquired in sorted order when multiple models are involved

## What changed

| File | Change |
|------|--------|
| `datastore_sync_coordinator.ts` | Refactored to `Map<string, SyncEntry>` with named lock support |
| `repo_context.ts` | Added `requireInitializedRepoUnlocked()`, `createModelLock()`, `acquireModelLocks()` |
| `s3_cache_sync.ts` | Added `pullChangedForModel()` for model-scoped S3 sync |
| `model_method_run.ts` | Uses per-model lock instead of global lock |
| `workflow_run.ts` | Extracts model refs, acquires per-model locks (falls back to global for dynamic refs) |
| `model_evaluate.ts` | Single-model evaluate uses per-model lock; `--all` keeps global |
| `workflow_evaluate.ts` | Single-workflow evaluate uses per-model lock; `--all` keeps global |
| `datastore_lock.ts` | `lock status` scans per-model locks; `lock release --model` added |
| `datastore_output.ts` | Shows lock scope (global vs per-model) in status output |
| `model_reference_extractor.ts` | New: extracts model references from workflows for lock targeting |

## Test plan

- [x] All 2968 existing tests pass
- [x] `deno check`, `deno lint`, `deno fmt` clean
- [x] New unit tests for coordinator, repo context helpers, and model reference extractor
- [x] Manual verification: two concurrent `model method run` on different models both proceed without blocking, each with its own `.lock` file
- [x] Binary compiles successfully

### DIFF
--- a/src/cli/repo_context.ts
+++ b/src/cli/repo_context.ts
@@ -619,38 +619,40 @@ export async function acquireModelLocks(
   }
 
   return async () => {
-    // For S3 datastores: push changes BEFORE releasing per-model locks.
-    // This prevents another process from acquiring the same per-model lock,
-    // pulling stale data from S3, and overwriting our local changes.
-    if (s3SyncService && config.type === "s3") {
-      const pushLock = createDatastoreLock(config);
-      try {
-        await pushLock.acquire();
-        logger.info`Pushing changes to S3...`;
-        const pushed = await s3SyncService.pushChanged();
-        if (pushed > 0) {
-          logger.info`Pushed ${pushed} file(s) to S3`;
-        }
-      } catch (error) {
-        const msg = error instanceof Error ? error.message : String(error);
-        logger.error("Failed to push changes to S3: {error}", {
-          error: msg,
-        });
-        throw new Error(
-          `S3 sync failed: could not push changes: ${msg}`,
-        );
-      } finally {
+    try {
+      // For S3 datastores: push changes BEFORE releasing per-model locks.
+      // This prevents another process from acquiring the same per-model lock,
+      // pulling stale data from S3, and overwriting our local changes.
+      if (s3SyncService && config.type === "s3") {
+        const pushLock = createDatastoreLock(config);
         try {
-          await pushLock.release();
-        } catch {
-          // Best-effort release
+          await pushLock.acquire();
+          logger.info`Pushing changes to S3...`;
+          const pushed = await s3SyncService.pushChanged();
+          if (pushed > 0) {
+            logger.info`Pushed ${pushed} file(s) to S3`;
+          }
+        } catch (error) {
+          const msg = error instanceof Error ? error.message : String(error);
+          logger.error("Failed to push changes to S3: {error}", {
+            error: msg,
+          });
+          throw new Error(
+            `S3 sync failed: could not push changes: ${msg}`,
+          );
+        } finally {
+          try {
+            await pushLock.release();
+          } catch {
+            // Best-effort release
+          }
         }
       }
-    }
-
-    // Release per-model locks AFTER push completes
-    for (const key of lockKeys) {
-      await flushDatastoreSyncNamed(key);
+    } finally {
+      // Always release per-model locks, even if S3 push fails
+      for (const key of lockKeys) {
+        await flushDatastoreSyncNamed(key);
+      }
     }
   };
 }


stack72:
## Manual testing of compiled binary

Tested the compiled binary against a fresh filesystem datastore to verify per-model locks work end-to-end.

### Setup

```bash
mkdir /tmp/swamp-lock-test
cd /tmp/swamp-lock-test
swamp repo init
```

Created two `command/shell` models with a 5-second sleep:

```yaml
# models/command/shell/model-a.yaml
type: "command/shell"
typeVersion: 1
id: "a0000000-0000-4000-a000-000000000001"
name: model-a
version: 1
tags: {}
globalArguments: {}
methods:
  execute:
    arguments:
      run: "sleep 5"
```

```yaml
# models/command/shell/model-b.yaml
type: "command/shell"
typeVersion: 1
id: "b0000000-0000-4000-b000-000000000002"
name: model-b
version: 1
tags: {}
globalArguments: {}
methods:
  execute:
    arguments:
      run: "sleep 5"
```

### Concurrent execution test

Launched both model method runs in the background simultaneously:

```bash
swamp model method run model-a execute &
swamp model method run model-b execute &
```

After 2 seconds (while both were still running), checked for lock files:

```
$ find .swamp -name ".lock"
.swamp/data/command/shell/a0000000-0000-4000-a000-000000000001/.lock
.swamp/data/command/shell/b0000000-0000-4000-b000-000000000002/.lock
```

**Two separate per-model lock files** — one per model, no global lock.

### Results

Both processes started at the same time and completed at the same time — no blocking:

```
# model-a
00:16:47.540 INF model·method·run·model-a·execute Found model "model-a" ("command/shell")
00:16:47.545 INF model·method·run·model-a·execute Executing method "execute"
00:16:52.729 INF model·method·run·model-a·execute Method executed
00:16:52.733 INF model·method·run·model-a·execute Method "execute" completed on "model-a"

# model-b
00:16:47.540 INF model·method·run·model-b·execute Found model "model-b" ("command/shell")
00:16:47.545 INF model·method·run·model-b·execute Executing method "execute"
00:16:52.723 INF model·method·run·model-b·execute Method executed
00:16:52.727 INF model·method·run·model-b·execute Method "execute" completed on "model-b"
```

Both started at `00:16:47.540` and finished at `~00:16:52.7` — running fully in parallel. With the old global lock, model-b would have waited ~5 seconds for model-a to release the lock before starting.

PR #722, opened 2026-03-16
fix: data GC never fully removes expired data from disk

### LINKED ISSUE #720: Data garbage collection never fully removes expired data from disk
## Description

Running `swamp data gc` reports expired data entries but never actually reclaims the disk space they occupy. Expired data accumulates on disk permanently because there is no code path that hard-deletes all versions of an expired data entry.

## Steps to Reproduce

1. Create model data with a short duration lifetime (e.g., `1m`)
2. Write several versions to the data
3. Wait for the lifetime to expire
4. Run `swamp data gc`
5. Observe that the data is reported as expired, but version directories remain on disk
6. Run `swamp data gc` again — the same data is reported as expired again

## Root Cause

The GC operates in two phases inside `deleteExpiredData()` in `data_lifecycle_service.ts`:

**Phase 1 — Lifecycle expiration (soft delete):** Calls `removeLatestMarker()` to delete the `latest` pointer file. No version directories are removed — zero disk space is freed.

**Phase 2 — Version GC:** Calls `collectGarbage()` on all models to trim excess versions based on the `garbageCollection` policy (e.g., keep N versions or versions within a duration).

There are two interacting problems:

### Problem 1: No hard-delete path for expired data

`collectGarbage()` always protects the latest version (`unified_data_repository.ts:1364`: `if (version !== Math.max(...versions))`). At least one version always survives. There is no code path that removes all versions of an expired entry.

For example, data with `garbageCollection: 5` and 3 versions that has expired:
- Phase 1 removes the `latest` marker (no disk space freed)
- Phase 2 sees 3 versions, policy is keep 5, `3 <= 5` — nothing to remove
- All 3 version directories remain on disk indefinitely

### Problem 2: Version GC undoes the soft-delete

When `collectGarbage()` does delete an old version, it calls `delete(type, modelId, dataName, version)` in `unified_data_repository.ts`. The `delete()` method (line 860-863) automatically recreates the `latest` marker pointing to the highest remaining version:

```
const versions = await this.listVersions(type, modelId, dataName);
if (versions.length > 0) {
  const newLatest = Math.max(...versions);
  await this.updateLatestMarker(type, modelId, dataName, newLatest);
}
```

This undoes the soft-delete from Phase 1. Additionally, `getLatestVersion()` has a fallback (line 1509-1511) that scans version directories when the `latest` file is missing, so even between phases the data remains accessible.

## Expected Behavior

`swamp data gc` should fully remove expired data entries from disk (all version directories), freeing the associated disk space. Running GC a second time should not find the same expired data again.

## Summary of Fix

The fix involves changes to the `deleteExpiredData` method in `data_lifecycle_service.ts` and potentially the `collectGarbage` or `delete` methods in `unified_data_repository.ts`:

- For expired data, all versions should be hard-deleted (not just the `latest` marker removed). This could be done by calling `delete(type, modelId, dataName)` without a version (which already removes the entire data directory) instead of `removeLatestMarker()`.
- If the two-phase approach is preserved, `collectGarbage` needs to be aware of soft-deleted data and skip recreating the `latest` marker for entries that have been expired. Alternatively, version GC should run before lifecycle expiration so the marker removal happens last.

## Environment

- swamp version: 20260315.010849.0-sha.5a9897f6
- Platform: macOS Darwin 25.3.0

stack72:
## Summary

Fixes #720

`swamp data gc` reported expired data but **never reclaimed any disk space**. The `deleteExpiredData()` method only called `removeLatestMarker()`, which removed the `latest` symlink but left all version directories and content files on disk.

This was a two-part problem:
1. **Phase 1** (expired data cleanup) only performed a soft-delete — no version directories were removed, zero bytes freed
2. **Phase 2** (`collectGarbage()`) always protects the latest version, and the single-version `delete()` path recreates the `latest` marker, effectively undoing Phase 1's soft-delete

## User impact

Before this fix, running `swamp data gc` would report expired entries but disk usage would remain unchanged. Users had no way to reclaim space from expired data without manually deleting directories from the datastore.

After this fix, `swamp data gc` fully removes expired data directories, correctly reports `versionsDeleted` and `bytesReclaimed`, and actually frees disk space.

## What changed

- **`data_lifecycle_service.ts`**: Replaced `removeLatestMarker()` (soft-delete) with `delete(type, modelId, dataName)` (hard-delete) in `deleteExpiredData()`. Before deleting, stats each version's content file to report accurate bytes reclaimed. Updated `LifecycleGCResult` doc comments to reflect the new behavior.

- **`data_lifecycle_service_test.ts`**: Added `delete` and `getContentPath` mocks to `MockDataRepository`. Added two new tests:
  - Verifies expired data triggers `delete()` (not `removeLatestMarker()`) and reports correct `versionsDeleted`
  - Verifies dry run does not call `delete()`

## Why this is correct

The no-version `delete(type, modelId, dataName)` path in `UnifiedDataRepository` does a clean `Deno.remove(dataNameDir, { recursive: true })` — it removes the entire data directory without recreating any markers. After Phase 1 hard-deletes expired entries, Phase 2's `findAllForModel()` naturally skips them since their directories no longer exist. No changes needed to `collectGarbage()` or the `delete()` implementation.

## Test plan

- [x] `deno check` — type checking passes
- [x] `deno lint` — no lint errors
- [x] `deno fmt` — properly formatted
- [x] `deno run test` — all 2997 tests pass (including 2 new tests)
- [x] `deno run compile` — binary compiles successfully

### DIFF
--- a/src/infrastructure/persistence/unified_data_repository.ts
+++ b/src/infrastructure/persistence/unified_data_repository.ts
@@ -1406,15 +1406,19 @@ export class FileSystemUnifiedDataRepository implements UnifiedDataRepository {
         }
       }
 
-      // Single latest marker update after all removals
+      // Re-scan actual versions after parallel deletions to avoid stale marker
       if (versionsToRemove.length > 0) {
-        const remaining = versions.filter((v) => !versionsToRemove.includes(v));
-        if (remaining.length > 0) {
+        const currentVersions = await this.listVersions(
+          type,
+          modelId,
+          data.name,
+        );
+        if (currentVersions.length > 0) {
           await this.updateLatestMarker(
             type,
             modelId,
             data.name,
-            Math.max(...remaining),
+            Math.max(...currentVersions),
           );
         } else {
           const dataNameDir = this.getDataNameDir(type, modelId, data.name);


keeb:
## Performance observation from real-world GC run

Tested this fix on a repo with ~4GB of accumulated data (game server metrics collected every 2 min via cron). The expired data phase works correctly — `delete()` with `rm -rf` is fast. However, the version GC phase in `collectGarbage` is very slow in practice.

### Root cause

In `collectGarbage`, old versions are removed one at a time (line 1371–1387):

```typescript
for (const version of versionsToRemove) {
  const stat = await Deno.stat(contentPath);   // 1 syscall per version
  await this.delete(type, modelId, data.name, version); // 1+ syscalls per version
  versionsRemoved++;
}
```

With terraria/minecraft metrics at version ~14,000+, this means ~14,000 individual stat + delete calls per data item. On NVMe the disk isn't the bottleneck — it's the per-call overhead from thousands of sequential async Deno syscalls.

### Additional overhead

- `findAllGlobal()` is called twice in `deleteExpiredData` — once in `findExpiredData()` and again at line 269 for the version GC pass
- `collectGarbage` calls `findAllForModel()` internally even though the caller just loaded all data

### Suggested fix direction

For the version GC path: when removing N consecutive old versions from a data-name directory, a single `Deno.remove(versionDir, { recursive: true })` per batch (or even `Deno.remove` on each version dir directly rather than going through the full `delete()` abstraction) would be much faster. The byte count can still be accumulated with a single `du`-style walk before deletion.

For the double scan: the second `findAllGlobal()` call could reuse the results already loaded in `findExpiredData()`, or both phases could be merged into one pass.

In our case GC went from 3.6GB → ~1GB over ~5 minutes on NVMe — functional but noticeably slow given the hardware.

keeb:
## Follow-up: tested the perf commit (d27c3227)

Recompiled and redeployed after `perf: optimize data GC with single scan and parallel batch deletion`. Results on the same repo:

- **First run (original fix commit):** ~5-10 min to clean 3.6GB → 2.4GB — functional but slow
- **Second run (perf commit):** cleaned 2.4GB → **12.5MB** very quickly

The single-scan + parallel batch deletion (concurrency=20, ~2 syscalls/version vs ~6) addressed exactly the bottlenecks I described. No further issues observed.

PR #849, opened 2026-03-24
feat: add `include` manifest field and loader pre-check for unbundled files

### LINKED ISSUE #707: Support unbundled helper scripts in extension packages
## Problem

Extension models that shell out to helper scripts (via `Deno.Command`) can't be published to the registry if the helper script imports npm packages with native/unbundleable dependencies.

The bundler tries to bundle every `.ts` file in `extensions/models/` and in the `additionalFiles` extraction path. If a helper script imports a package like `hap-controller` (which has native deps for Bluetooth that are never used at runtime), the bundle fails and the extension can't be pushed.

## Context

The `@bixu/homekit` extension discovers HomeKit accessories via mDNS using `hap-controller`. The main model (`homekit.ts`) bundles fine — it only imports `zod` and shells out to a helper script. The helper script (`homekit_discover.ts`) imports `hap-controller` and runs as a standalone `deno run` subprocess. This works locally in `extensions/scripts/` but can't be published because:

1. `additionalFiles` are extracted to `extensions/models/`, where swamp tries to bundle them as models
2. There's no way to include a file that skips bundling

## Proposed Solution

Support an `extensions/scripts/` directory (or similar) that:

- Is included in the extension archive during `push`
- Is extracted to a known, predictable path on `pull`
- Is **not** bundled or loaded as a model by swamp
- Can be referenced by models at runtime via `context.repoDir`

Alternatively, a `scripts` key in the manifest (alongside `models` and `workflows`) that marks files as "include but don't bundle":

```yaml
manifestVersion: 1
name: "@bixu/homekit"
version: "2026.03.14.1"
models:
  - homekit.ts
scripts:
  - homekit_discover.ts
```

## Current workaround

Keep the helper script in `extensions/scripts/` as a local-only file. The extension works in the repo but can't be published.

stack72:
## Summary

- Adds an `include` manifest field for shipping helper files alongside models without bundling them
- Adds a source pre-check to all five loaders to skip files that don't declare the expected export
- Fixes #707 where helper scripts with unbundleable native dependencies break extension loading

## Problem

Extension models that shell out to helper scripts (via `Deno.Command`) can't be published if the helper script imports packages with native/unbundleable dependencies. The loader discovers every `.ts` file in the models directory, tries to bundle each one, and fails on files with native deps — even though those files aren't models.

Example: the `@bixu/homekit` extension has a main model (`homekit.ts`) that bundles fine, and a helper script (`homekit_discover.ts`) that imports `hap-controller` (Bluetooth native deps). The model shells out to the helper via subprocess. Locally it works, but bundling fails and the extension can't be pushed or loaded after pull.

### Part 1: Loader pre-check (runtime fix)

All five loaders (model, vault, driver, datastore, report) now read the TypeScript source before attempting to bundle and check for the expected export pattern:

| Loader | Regex |
|---|---|
| Model | `export\s+const\s+(model\|extension)\s*[=:]` |
| Vault | `export\s+const\s+vault\s*[=:]` |
| Driver | `export\s+const\s+driver\s*[=:]` |
| Datastore | `export\s+const\s+datastore\s*[=:]` |
| Report | `export\s+const\s+report\s*[=:]` |

If the pattern isn't found, the file is skipped with a debug log — no bundling attempted, no error. The regex uses `[=:]` to handle both direct assignment (`export const model = {`) and type-annotated forms (`export const model: Type = {`).

**Why this is safe:**
- Every existing extension uses `export const model = { ... }` directly — zero cases of non-standard export patterns in the codebase or swamp-extensions repo
- Files that ARE models match the regex and proceed through the existing bundle → import → validate flow unchanged
- Files that aren't models (helpers, utilities) are skipped earlier than before — previously they'd be bundled, imported, and silently skipped at the export check; now they're skipped before bundling
- Debug log (`Skipping ${file} (no model/extension export found)`) surfaces skipped files for troubleshooting

**Performance:** Benchmarked at 0.057ms per file read + regex. With the `@swamp/aws/ec2` extension (many model files), no measurable difference between the new binary and the system binary (both ~3.5-4.8s, variance between runs exceeds any overhead).

### Part 2: `include` manifest field (publishing fix)

New optional manifest field that tells the push command "include these files in the archive without bundling them":

```yaml
manifestVersion: 1
name: "@bixu/homekit"
version: "2026.03.14.1"
models:
  - homekit.ts
include:
  - homekit_discover.ts
```

**Push flow:**
- Include files are resolved relative to `modelsDir` (same as models — preserves relative path references)
- Copied to the archive under `models/` (so they end up alongside model sources)
- NOT bundled — no entry in `bundles/`
- NOT quality-checked (they may have their own lint/fmt conventions)
- ARE safety-checked (scanned for disallowed patterns like eval, symlinks, etc.)
- Written to the archive manifest so pull can read them

**Pull flow:**
- Include files extract naturally from `models/` in the archive to `modelsDir`
- Recorded in `upstream_extensions.json` under a new `include` field
- The loader's pre-check skips them at runtime (they don't export `model`/`extension`)

### Manifest & resolution
- **`extension_manifest.ts`** — Added `include: string[]` to schema, interface, and parser
- **`resolve_extension_files.ts`** — Added `includeFilePaths` resolution (relative to `modelsDir`) and return field

### Loaders (pre-check)
- **`user_model_loader.ts`** — Pre-check for `export const (model|extension)`
- **`user_vault_loader.ts`** — Pre-check for `export const vault`
- **`user_driver_loader.ts`** — Pre-check for `export const driver`
- **`user_datastore_loader.ts`** — Pre-check for `export const datastore`
- **`user_report_loader.ts`** — Pre-check for `export const report`

### Push/pull
- **`push.ts`** (libswamp) — Accepts `includeFilePaths`, copies to archive, writes to manifest, excludes from quality checks, includes in safety analysis
- **`pull.ts`** (libswamp) — Reads manifest `include`, passes to `updateUpstreamExtensions`
- **`extension_push.ts`** (CLI) — Passes `includeFilePaths` through both push calls (initial + version-bump retry)
- **`upstream_extensions.ts`** — Added `include?: string[]` to entry type
- **`push_test.ts`** — Updated test helpers with `include` field

## Design decisions

1. **Pre-check regex, not manifest-driven skip**: The loader pre-check works for both local extensions (no manifest) and pulled extensions. No infrastructure needed — the source itself is the signal.

2. **`[=:]` not just `=`**: The regex matches type-annotated exports (`export const model: Type =`) in addition to direct assignment. Zero existing extensions use type annotations, but this prevents a future surprise where a model silently stops loading.

3. **Include files in `models/` not a new directory**: Include files sit alongside model sources in the archive and on disk. This preserves relative path references — if a model shells out to `./homekit_discover.ts`, the file is right there.

4. **Quality checks excluded, safety included**: Include files may have their own lint/fmt conventions (or be deliberately un-linted). But safety analysis still scans them for disallowed patterns — they're still distributed code.

5. **`upstream_extensions.json` for tracking, not loader skip**: The `include` field in `upstream_extensions.json` tracks which files are include files for removal/update purposes. The loader doesn't read it — the pre-check handles skipping. This avoids coupling the loader to the upstream tracking file.

### Automated (3549 tests, 0 failed)
- `deno check` — zero type errors
- `deno lint` — all files clean
- `deno fmt --check` — all files clean
- `deno run test` — 3549 tests passed, 0 failed (run twice to confirm no flaky failures)

### Manual end-to-end

**1. Loader pre-check — helper script skipped**
- Created extension with `homekit.ts` (model) and `homekit_discover.ts` (helper, no model export)
- `swamp model type search homekit` — model loaded, helper skipped
- `.swamp/bundles/` contains only `homekit.js` — helper was not bundled

**2. Push with `include` field**
- Created manifest with `include: [homekit_discover.ts]`
- `swamp extension push manifest.yaml --dry-run` — succeeded, 1.6KB archive
- Helper was NOT quality-checked (no lint errors from helper)
- Helper WAS included in safety analysis

**3. Pull + load existing extension (regression)**
- `swamp extension pull @stack72/letsencrypt-certificate@2026.03.04.1` — succeeded
- `swamp model type search letsencrypt` — found and loaded

**4. Inline deps push (regression)**
- Extension with `npm:zod@4.3.6` imports, no config files
- `swamp extension push manifest.yaml --dry-run` — 203.5KB archive, zod externalized

**5. Performance benchmark (no regression)**
- `@swamp/aws/ec2` extension (many model files)
- New binary: 3644-4239ms across 3 runs
- System binary: 3483-4811ms across 3 runs
- No measurable difference — variance between runs exceeds any pre-check overhead

## Test plan

- [ ] CI passes (type check, lint, fmt, tests)
- [ ] Verify helper files alongside models are skipped by loader (no bundle error)
- [ ] Verify `include` files are in archive but not bundled during push
- [ ] Verify existing extensions pull and load correctly (regression)
- [ ] Verify `@swamp/aws/ec2` loads at same speed as before

Closes #707


stack72:
## Detailed testing results

### Loader pre-check verification

Created an extension with two files in `extensions/models/`:
- `homekit.ts` — has `export const model = { ... }`, should be bundled
- `homekit_discover.ts` — helper script, no `export const model`, should be skipped

**Result:** `swamp model type search homekit` found `@test/homekit`. Only `homekit.js` appeared in `.swamp/bundles/` — no `homekit_discover.js`. The helper was silently skipped without error.

### Push with `include` field

Manifest:
```yaml
manifestVersion: 1
name: "@test/homekit"
version: "2026.03.24.1"
models:
  - homekit.ts
include:
  - homekit_discover.ts
```

**Result:** `swamp extension push manifest.yaml --dry-run` produced 1.6KB archive. The include file was NOT quality-checked (previously it failed with lint errors from the helper). Safety analysis still ran on both files.

### Quality check exclusion verified

Before fixing the quality check separation, the push failed with lint errors from the helper file (`require-await` on helper functions). After separating `qualityFiles` from `allFiles`, the push succeeded — helper file excluded from lint/fmt but still included in safety analysis.

### Regression: Pull + load existing extension

```
swamp extension pull @stack72/letsencrypt-certificate@2026.03.04.1
swamp model type search letsencrypt
```

**Result:** Pulled successfully, model loaded and found in search.

### Regression: Inline deps push

Extension using `npm:zod@4.3.6` inline imports (no deno.json, no package.json):

**Result:** `swamp extension push manifest.yaml --dry-run` produced 203.5KB archive with zod externalized. Identical to previous behaviour.

### Performance benchmark: `@swamp/aws/ec2` extension

Pulled `@swamp/aws/ec2` (many model files) and ran `swamp model type search ec2` three times with each binary:

| | Run 1 | Run 2 | Run 3 |
|---|---|---|---|
| **New binary** (with pre-check) | 3725ms | 4239ms | 3644ms |
| **System binary** (without pre-check) | 4811ms | 3624ms | 3483ms |

**No measurable difference.** The variance between runs (~500-1000ms) is larger than any overhead from the pre-check. The file read + regex check benchmarked at 0.057ms per file — negligible compared to the 30-100ms per file for bundling.

### Regex safety verification

Verified the pre-check regex `export\s+const\s+(model|extension)\s*[=:]` against all patterns:

| Source pattern | Matches? | Correct? |
|---|---|---|
| `export const model = {` | Yes | ✅ Expected |
| `export const model: Type = {` | Yes | ✅ Expected (type annotation) |
| `export const extension = {` | Yes | ✅ Expected |
| `export function discover()` | No | ✅ Expected (helper) |
| `const model = { ... }` (no export) | No | ✅ Expected (internal) |
| `export { model } from "./impl"` | No | ⚠️ Theoretical risk — no existing extensions use this pattern |

Searched the entire codebase and swamp-extensions repo for non-standard export patterns — zero matches found. All existing extensions use `export const model = { ... }` directly.

Issue #902, opened 2026-03-28
vault.get() regex does not support spaces in vault names or secret keys


danmcclain:
## Description

The `vault.get()` regex in `src/domain/expressions/model_resolver.ts` does not support spaces in vault names or secret keys, even when quoted.

## Steps to Reproduce

1. Create a vault backed by 1Password: `swamp vault create @swamp/1password infra --config '{"op_vault": "Infra"}'`
2. The 1Password item has fields with spaces, e.g. `Client ID` and `Client Secret`
3. Use a vault expression in a workflow: `${{ vault.get(infra, Tailscale K8s Operator/Client ID) }}`
4. Run the workflow

**Expected:** The vault expression resolves the secret value.

**Actual:** CEL parse error: `Expected RPAREN, got IDENTIFIER` — the space in the key breaks parsing before the vault regex even runs.

Even if the CEL layer is bypassed, the vault regex itself would fail:

```typescript
const vaultPattern = /vault\.get\(\s*(['"\`]?)([^'"\`\s,]+)\1\s*,\s*(['"\`]?)([^'"\`\s,]+)\3\s*\)/g;
```

The capture group `[^'"\`\s,]+` explicitly excludes whitespace, so `Tailscale K8s Operator/Client ID` stops matching at the first space.

## Workaround

Rename the 1Password fields to remove spaces (e.g. `ClientID` instead of `Client ID`), or use the item UUID with spaceless field names.

## Suggested Fix

Update the vault regex to allow spaces inside quoted arguments. When quotes are present, the inner capture should permit any character except the matching quote. For example:

```
/vault\.get\(\s*(?:(['"\`])(.+?)\1|([^\s,)]+))\s*,\s*(?:(['"\`])(.+?)\4|([^\s,)]+))\s*\)/g
```

This would allow both `vault.get(infra, mykey)` and `vault.get("infra", "My Key With Spaces")`.

The CEL expression parser layer (`${{ }}`) would also need to handle the vault.get() call before passing to the CEL evaluator, since CEL itself tokenizes on spaces.

## Environment

- swamp version: 20260327.235335.0-sha.0c993a65
- OS: macOS

keeb:
🔍 **Triage started** — fetching issue context

keeb:
[...]

PR #905, opened 2026-03-28
fix: quote --allowedTools in issue-triage workflow

### LINKED ISSUE #904: fix: issue-triage workflow fails due to shell-quote parsing of --allowedTools
## Description

The `/triage` command fails on all issues since #852 hardened the `--allowedTools` patterns.

## Root Cause

`claude-code-action` uses `shell-quote` to parse `claude_args`. The unquoted `Bash(...)` patterns containing spaces get mangled:

1. Parentheses `()` are treated as shell operators and stripped
2. Spaces inside patterns cause word-splitting — e.g. `Bash(gh api --method POST:*/reactions)` becomes separate tokens
3. `--method` from inside a Bash pattern gets parsed as a standalone CLI flag, corrupting the argument structure
4. Claude Code exits with code 1

The workflow logs from #902's triage attempt confirm the allowed tools were parsed as `["Read", "Glob", "Grep", "Bash", "gh", "issue", "api"]` instead of the intended `Bash(...)` patterns.

## Fix

Wrap the `--allowedTools` value in single quotes so `shell-quote` preserves it as a single token.

## Introduced In

#852 (commit 8eab7fe6)

keeb:
## Summary

- Fix `/triage` command failure caused by `shell-quote` parsing of unquoted `Bash(...)` patterns in `claude_args`
- Wraps the `--allowedTools` value in single quotes so spaces and parentheses are preserved as a single token

## Root Cause

`claude-code-action` uses `shell-quote` to parse `claude_args`. The `Bash(...)` patterns added in #852 contain spaces (e.g. `Bash(gh api --method POST:*/reactions)`), which caused:
1. Parentheses treated as shell operators and stripped
2. Space-splitting broke patterns into separate tokens
3. `--method` parsed as a standalone CLI flag, corrupting the entire argument structure
4. Claude Code crashed with exit code 1

Confirmed by workflow logs from the failed #902 triage — allowed tools were `["Read", "Glob", "Grep", "Bash", "gh", "issue", "api"]` instead of the intended patterns.

Fixes #904

## Test plan

- [ ] Re-run `/triage` on #902 after merge and verify it completes successfully
- [ ] Verify the workflow logs show the correct `Bash(...)` patterns in `allowedTools`

### DIFF
--- a/.github/workflows/issue-triage.yml
+++ b/.github/workflows/issue-triage.yml
@@ -161,7 +161,7 @@ jobs:
           anthropic_api_key: ${{ secrets.ANTHROPIC_API_KEY }}
           claude_args: |
             --model claude-opus-4-5-20251101
-            --allowedTools Read,Glob,Grep,Bash(gh issue view:*),Bash(gh issue comment:*),Bash(gh api --method POST:*/reactions),Bash(gh api --method PATCH:*/comments/*)
+            --allowedTools 'Read,Glob,Grep,Bash(gh issue view:*),Bash(gh issue comment:*),Bash(gh api --method POST:*/reactions),Bash(gh api --method PATCH:*/comments/*)'
 
       - name: Add completion reaction
         if: success()


stack72:
@bixu thanks for filing this — it's a thoughtful observation about the gap.

After digging into it, we're leaning toward not adding a test coverage factor to the rubric. The core concern is false-positive risk: test *presence* doesn't correlate with test *quality* the way the other rubric factors do. A README being substantive (≥500 chars, ≥2 code blocks) is a reasonable proxy for documentation quality. But a single `Deno.test("noop", () => {})` would earn full points on a test factor, and once points are on the table, there's an incentive to add the minimum rather than write meaningful tests. That would actively erode trust in the overall score.

A couple of things that already partially cover this:

- There's a `testing-completeness` review rule that warns extension authors at push time when entry points lack sibling `_test.ts` files — so authors do get nudged.
- The `repository-verified` rubric factor links consumers to the source repo, where they can inspect tests, CI, and coverage directly.

Shipping test files in the bundle (your second proposal) would add download size for something the verified repo link already provides, and still wouldn't tell you whether the tests are meaningful.

We think the right answer here is: if you're evaluating whether to trust a third-party extension's correctness, the source repository is the place to do that — and the rubric already verifies and links to it. Adding a score factor that can be trivially gamed would give false confidence rather than real signal.

Ping me if you want to talk about this more. Appreciate you raising it — closing this one out.

PR #919, opened 2026-03-30
feat: add context.readModelData() for first-class cross-model data access

### LINKED ISSUE #914: context.readModelData(modelName) — first-class cross-model data access in execute functions
## Problem

Extension models that need to read another model's data at runtime have no first-class API to do so. The `dataRepository.findAllForModel(type, modelId)` method requires a type object (not a string), and there's no way to resolve a model name to its type and ID from within an execute function.

Currently, models that compose with other models' output (e.g., a dedup filter reading from a source model) must:

1. Scan YAML files on the filesystem to resolve model name → type + ID
2. Read raw JSON files from `.swamp/data/{type}/{id}/{name}/{version}/raw`
3. Parse the version pointer from the `latest` file manually

This works but is fragile, undocumented, and reimplements what the workflow engine already does internally.

## Proposed Solution

Add a `context.readModelData(modelName, specName?)` API to the execute function context that:

1. Resolves a model name to its type and ID using the definition repository
2. Returns all data items for that model (or filtered by spec name)
3. Returns the parsed data content, not just metadata

```typescript
// Current workaround (fragile)
const resolved = await resolveModelByName(context.repoDir, "anime-source");
const dataDir = `${context.repoDir}/.swamp/data/${resolved.type}/${resolved.id}`;
// ... manual filesystem traversal ...

// Proposed API
const episodes = await context.readModelData("anime-source", "episode");
// Returns: Array<{ name: string, data: Record<string, unknown> }>
```

The `definitionRepository` is available on report contexts but not on execute contexts, and even there it lacks a `getByName()` method.

## Scope

Changes needed in:
- Execute context construction: expose `definitionRepository` or a model resolver
- `dataRepository`: add a name-based lookup that resolves model name → type + ID internally
- Possibly a new convenience method on context that combines resolution + data reading

This would make model composition a first-class pattern rather than a filesystem hack.

## Use Case

A dedup/filter model reads episode data from a source model, checks local directories, and outputs only new episodes. The source model is specified by name at runtime. Without this API, the dedup model contains ~80 lines of YAML-scanning and filesystem-reading code just to access another model's data.

keeb:
## Summary

- Adds `context.readModelData(modelName, specName?)` to `MethodContext` so extension models can read another model's data without manual definition resolution and raw repository calls
- Introduces `DataAccessService` domain service that handles model name resolution, data reading, JSON parsing, vault ref resolution, and orphan data recovery
- Extracts `DataRecord` value object to `src/domain/data/data_record.ts` as a shared type between the expression system and data access layer

## Test Plan

- 10 unit tests for `DataAccessService` covering: model resolution, spec filtering, empty/missing models, renamed data, vault refs, non-JSON content, orphan recovery
- 4 integration tests with real file-backed repositories: cross-model read, spec filter, non-existent model, no-data model
- All 3699 existing tests pass
- `deno check`, `deno lint`, `deno fmt` clean

Closes #914

### DIFF
--- a/.claude/skills/swamp-data/SKILL.md
+++ b/.claude/skills/swamp-data/SKILL.md
@@ -5,29 +5,30 @@ description: List model data, view version history, delete expired versions, and
 
 # Swamp Data Skill
 
-Manage model data lifecycle through the CLI.
+Manage model data lifecycle through the CLI. All commands support `--json` for
+machine-readable output.
 
 **Verify CLI syntax:** If unsure about exact flags or subcommands, run
 `swamp help data` for the complete, up-to-date CLI schema.
 
 ## Quick Reference
 
-| Task                   | Command                                        |
-| ---------------------- | ---------------------------------------------- |
-| Search all data        | `swamp data search`                            |
-| Search with filters    | `swamp data search --type output --since 1d`   |
-| Search by workflow     | `swamp data search --workflow my-workflow`     |
-| Search by model        | `swamp data search --model my-model`           |
-| Free-text search       | `swamp data search vpc`                        |
-| List model data        | `swamp data list <model>`                      |
-| List workflow data     | `swamp data list --workflow <name>`            |
-| Get specific data      | `swamp data get <model> <name>`                |
-| Get metadata only      | `swamp data get <model> <name> --no-content`   |
-| Get data via workflow  | `swamp data get --workflow <name> <data_name>` |
-| View version history   | `swamp data versions <model> <name>`           |
-| Run garbage collection | `swamp data gc`                                |
-| Rename data instance   | `swamp data rename <model> <old> <new>`        |
-| Preview GC (dry run)   | `swamp data gc --dry-run`                      |
+| Task                   | Command                                               |
+| ---------------------- | ----------------------------------------------------- |
+| Search all data        | `swamp data search --json`                            |
+| Search with filters    | `swamp data search --type output --since 1d --json`   |
+| Search by workflow     | `swamp data search --workflow my-workflow --json`     |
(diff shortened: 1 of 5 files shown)


keeb:
## First consumer report

Used this in the `@keeb/mms-dedup` extension model tonight. The dedup model reads episode data from a source model (by name), checks local directories for existing files, and outputs only new episodes.

### Before (workaround)

~80 lines of filesystem hacking:
- `resolveModelByName()` — scan YAML files in `models/` to find type + ID
- `readLatestData()` — walk `.swamp/data/{type}/{id}/`, read `latest` pointer files, parse `raw` JSON
- `walkYaml()` — recursive directory walker

```typescript
const resolved = await resolveModelByName(context.repoDir, args.sourceModel);
const dataDir = `${context.repoDir}/.swamp/data/${resolved.type}/${resolved.id}`;
const items = await readLatestData(dataDir);
```

### After

```typescript
const items = await context.readModelData(args.sourceModel, "episode");
```

One line. Model name resolution, data reading, JSON parsing — all handled.

### On coupling concerns

The API is read-only and name-based, which keeps it loose. The dedup model doesn't know or care about the source model's type, schema, or internals — it just reads data by name and spec. The coupling is the same as what CEL expressions already do in workflows (`data.findBySpec("anime-source", "episode")`), just available inside execute functions.

The alternative was each model being an island at runtime, which pushed all composition into workflow YAML. That works for simple chains but falls apart when a model needs to iterate over another model's data with logic (LLM-parse if raw, skip if pre-parsed, deduplicate within results). Workflows can't express that — it requires code.

The key property: models can read each other's data but can't write to each other's namespace. That's the right boundary.

PR #933, opened 2026-03-30
fix: prevent orphaned lock on Ctrl-C by closing heartbeat/release race

### LINKED ISSUE #928: Lock not released when extension pull is interrupted (Ctrl-C)
## Description

When running `swamp extension pull` and interrupting it mid-way (e.g. Ctrl-C or process kill), the datastore lock is not released. On the next invocation of any write command, swamp hangs waiting to acquire the lock rather than detecting the previous holder's process is dead and auto-expiring.

## Steps to Reproduce

1. Run `swamp extension pull @some/extension`
2. Interrupt with Ctrl-C before it completes
3. Run any write command (e.g. `swamp extension pull` again, `swamp model method run`, etc.)
4. Command hangs indefinitely

## Expected Behavior

The lock should auto-expire after its TTL (30s) since the previous process is dead, or swamp should detect that the lock holder PID is no longer running and release it automatically.

## Actual Behavior

The lock remains held and the next command hangs until manually released with `swamp datastore lock release --force`.

## Workaround

```
swamp datastore lock release --force
```

## Environment

- swamp version: [1mswamp[22m [94m20260329.231615.0-sha.0e8efc6b[39m
- OS: macOS
- Datastore: filesystem

stack72:
## Summary

Fixes #928 — lock not released when `extension pull` (or any command) is interrupted with Ctrl-C.

**Root cause:** Race condition between the heartbeat `extend()` method and `release()` in `FileLock`/`S3Lock`. When Ctrl-C fires, the SIGINT handler calls `release()` which deletes the lock file, but if `extend()` was mid-flight it would recreate the lock file after deletion. `Deno.exit(130)` then runs before `extend()`'s post-write cleanup can execute, leaving an orphaned lock.

**Fix:**
- Add a `releasing` flag checked by `extend()` between its async read and write — if set, `extend()` returns without writing, preventing the orphan
- Add PID-based stale lock detection (`Deno.kill(pid, "SIGCONT")`) to `FileLock` — when a lock holder's process is dead, the lock is immediately reclaimed instead of waiting ~30s for TTL expiry

## Test plan

- [x] New test: stale lock from dead process is immediately acquired (FileLock)
- [x] New test: lock held by live process is not stolen via PID check (FileLock)
- [x] New test: release cleans up lock even during heartbeat (S3Lock)
- [x] All 3733 existing tests pass
- [x] `deno check`, `deno lint`, `deno fmt` all pass

### DIFF
--- a/src/infrastructure/persistence/file_lock.ts
+++ b/src/infrastructure/persistence/file_lock.ts
@@ -36,6 +36,25 @@ import type {
 import { LockTimeoutError } from "../../domain/datastore/distributed_lock.ts";
 import { getSwampLogger } from "../logging/logger.ts";
 
+/**
+ * Check if a process with the given PID is no longer running.
+ * Uses signal 0 (no-op) to test process existence.
+ * Returns false (not dead) on any error other than NotFound,
+ * so TTL-based detection remains the fallback.
+ */
+function isProcessDead(pid: number): boolean {
+  try {
+    Deno.kill(pid, "SIGCONT");
+    return false; // Process exists
+  } catch (error) {
+    if (error instanceof Deno.errors.NotFound) {
+      return true; // Process does not exist
+    }
+    // PermissionDenied or other error — can't determine, fall back to TTL
+    return false;
+  }
+}
+
 const DEFAULT_TTL_MS = 30_000;
 const DEFAULT_RETRY_INTERVAL_MS = 1_000;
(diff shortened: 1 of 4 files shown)


stack72:
Closing in favor of a rebased PR — main removed the built-in S3Lock (moved to @swamp/s3-datastore extension).

PR #1035, opened 2026-04-01
fix: address adversarial review findings for data query

adamhjk:
## Summary

Fixes three medium-severity findings from the adversarial review on #1029:

- **Paged iteration**: `CatalogStore.iterate()` now fetches rows in batches of 1000 via `LIMIT/OFFSET` instead of `stmt.all()` which loaded all rows into memory
- **Serve shutdown cleanup**: Close `CatalogStore` after `server.finished` so the SQLite connection isn't leaked for long-running processes
- **Debug logging**: Log skipped rows at debug level in `DataQueryService` instead of silently swallowing evaluation errors, visible with `--verbose`

Also incorporates UX review suggestions from the same PR: CLI examples, improved `--select` description, and JSON envelope for projected output.

## Test Plan

- All 3940 existing tests pass
- `deno fmt`, `deno lint`, `deno check` all clean
- Manual: `swamp data query 'size > 0' --verbose` shows debug output for type-mismatch rows

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/src/cli/commands/data_query.ts
+++ b/src/cli/commands/data_query.ts
@@ -39,7 +39,19 @@ export const dataQueryCommand = new Command()
   .arguments("<predicate:string>")
   .option("--repo-dir <dir:string>", "Repository directory", { default: "." })
   .option("--limit <n:number>", "Maximum results", { default: 100 })
-  .option("--select <expr:string>", "CEL projection expression")
+  .option(
+    "--select <expr:string>",
+    "CEL expression to extract fields from matching records (e.g. data.name)",
+  )
+  .example("Filter by model", "swamp data query 'modelName == \"scanner\"'")
+  .example(
+    "Filter with size threshold",
+    "swamp data query 'size > 1048576'",
+  )
+  .example(
+    "Project a single field",
+    "swamp data query 'dataType == \"resource\"' --select data.name",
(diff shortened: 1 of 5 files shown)


stack72:
@dmc Thanks for the detailed report — the analysis of the contract gap was really well done.

After reviewing the codebase, we scoped the fix to the silent failure on invalid verdict values (your point #3). When a report contained an unrecognized verdict like `"fail"`, the Zod validation silently discarded the entire report and re-emitted "No adversarial review recorded" — as if you'd never written one. That's now an explicit error that names the offending dimension and lists the allowed values (`pass`, `issue`, `na`, `pending`).

For points #1 and #2 (embedding dimension definitions and documenting the enum in the skeleton), we decided not to act on those at this time. The skeleton is designed to be consumed by Claude Code agents rather than read directly by humans — agents have the full skill reference (including all dimension definitions and verdict values) already in their context when they fill it in. Enriching the skeleton with inline definitions would be redundant for the intended consumer. The DX gap you hit was more about the agent not cross-referencing those docs properly than missing information in the skeleton itself.

Ping @stack72 on discord directly if you want to talk this over.

Fix is in #1806.

PR #1112, opened 2026-04-06
fix: anchor catalog type extraction regex to export block

### LINKED ISSUE #1110: issue-lifecycle skill fails in worktrees: missing init and dev extension model resolution
## Problem

The `/issue-lifecycle` skill is brittle when running in Claude Code worktrees. It sometimes works, sometimes doesn't. Two distinct failure scenarios:

### 1. Worktree not initialized
When Claude Code creates a worktree via `EnterWorktree`, it creates a git worktree but doesn't run `swamp repo init`. Without `.swamp/` and `.swamp.yaml`, all `swamp model` commands fail. The skill doesn't include initialization as a prerequisite step.

### 2. Scratch repo can't load dev extension models  
During triage reproduction (Phase 1 Step 5), when a scratch repo is created at `/tmp/swamp-repro-issue-<N>`, dev extension models like `@si/issue-lifecycle` aren't available. The Deno bundler can't resolve zod imports because the scratch repo has no `deno.json` import map. The workaround is `SWAMP_MODELS_DIR=<worktree>/extensions/models`, but this isn't documented in the skill.

### 3. `findNearestDenoConfig` boundary check ordering
In `user_model_loader.ts:1084-1105`, the repo boundary check fires before the deno.json check, meaning if a model file is at the repo root level, its deno.json would be missed.

## Expected Behavior

The issue-lifecycle skill should work flawlessly every time in any context (main repo, worktree, scratch repo) without manual environment setup.

## Proposed Fix

1. **Skill**: Add Phase 0 (Environment Setup) with worktree detection, `swamp repo init`, model verification, and `SWAMP_MODELS_DIR` guidance
2. **Skill**: Update triage.md scratch repo instructions to set `SWAMP_MODELS_DIR`
3. **Code**: Fix `findNearestDenoConfig` boundary check order
4. **Code**: Improve error messages when model type not found

keeb:
## Summary

Fixes #1110

- **Root cause**: The catalog population regex `/type\s*:\s*["']([^"']+)["']/` matched the first `type: "..."` anywhere in a model source file. In `issue_lifecycle.ts`, `type: "feature"` on line 81 (inside a `sc.ensureIssue()` call) was matched before the actual `type: "@si/issue-lifecycle"` on line 113, permanently poisoning the extension catalog.
- **Fix**: Anchor both type and version regexes to the `export const (model|extension) = {` block using `/export\s+const\s+(?:model|extension)\s*=\s*\{[\s\S]*?type\s*:\s*["']([^"']+)["']/`
- **Test**: Added regression test with a decoy `type:` property before the export block
- **Skill**: Added Phase 0 (Environment Setup) to issue-lifecycle skill for worktree init and `SWAMP_MODELS_DIR` guidance in scratch repos

## Test Plan

- [x] New unit test: `catalog extracts type from export block, not from decoy properties` passes
- [x] All 484 model domain tests pass
- [x] `deno check` passes
- [x] `deno lint` clean (pre-existing worktree issues only)
- [x] `deno fmt` clean
- [x] Manual verification: cleared catalog, ran `model create @si/issue-lifecycle`, confirmed catalog stores `@si/issue-lifecycle` (not `feature`), subsequent commands work without clearing catalog

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -162,7 +162,10 @@ jobs:
       - name: Setup Node.js
         uses: actions/setup-node@v6
         with:
-          node-version: "24"
+          # Pinned to 20 LTS. Node 22+ rejects require() of ESM modules
+          # with top-level await (ERR_REQUIRE_ASYNC_MODULE), which breaks
+          # promptfoo's transitive dep @asamuzakjp/css-color via jsdom.
+          node-version: "20"
 
       - name: Run skill trigger evals
         env:
--- a/scripts/eval_skill_triggers_promptfoo.ts
+++ b/scripts/eval_skill_triggers_promptfoo.ts
@@ -178,59 +178,11 @@ async function main(): Promise<void> {
     `Running skill trigger evals for ${model} (concurrency=${concurrency}, threshold=${passThreshold})…`,
   );
 
-  // Install promptfoo locally with an override to pin
-  // @asamuzakjp/css-color to 5.1.4. Version 5.1.5+ uses top-level await
-  // in its ESM entry, which Node cannot require() (ERR_REQUIRE_ASYNC_MODULE).
-  // This is a transitive dep: promptfoo → jsdom → @asamuzakjp/css-color.
-  // We use `npm install` instead of `npx -y` so overrides take effect.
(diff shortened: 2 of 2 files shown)


stack72:
superceeded by #1121 & #1115

Issue #947, opened 2026-03-30
Workflow step execution bypasses modelMethodRun, diverging from direct method execution


keeb:
## Description

Workflow step execution and direct `swamp model method run` use different code paths to execute model methods. Direct execution goes through `modelMethodRun()` in `src/libswamp/models/run.ts`, while workflow execution has its own method invocation path in `src/libswamp/workflows/`.

This means features added to `modelMethodRun` (report context fields like `swampVersion`, `outputSpecs`, input validation behaviour, etc.) don't automatically apply to workflow-triggered method runs. Each enhancement has to be wired in twice, and divergence accumulates silently.

Discovered while adding `swampVersion` to the method-summary report context — it was straightforward to pass from the CLI command through `ModelMethodRunInput`, but workflow-triggered runs have no equivalent path, so the version is absent from their reports.

## Steps to Reproduce

1. Add a new field to `ModelMethodRunInput` and thread it into `MethodReportContext`
2. Pass it from `src/cli/commands/model_method_run.ts`
3. Observe that workflow-triggered method executions in `src/libswamp/workflows/` don't go through the same path and lack the new field

## Expected Behaviour

A single method execution path that both direct CLI invocation and workflow step execution use, so enhancements apply uniformly.

## Environment

swamp 20260206.200442.0

stack72:
🔍 **Triage started** — fetching issue context

stack72:
📋 **Classified as bug** (high) — Confirmed: workflow step execution in execution_service.ts:182-740 has its own method invocation path that bypasses modelMethodRun in libswamp/models/run.ts:218-785. Key divergences: (1) no failed-execution reports in workflow path, (2) missing swampSha and outputSpecs in report context, (3) no env var detection warning, (4) different input validation (no effective schema building), (5) check skip options not forwarded. Both paths converge at executionService.executeWorkflow() but with different MethodContext populations.

stack72:
[...]

PR #1164, opened 2026-04-11
fix: consolidate MethodContext construction through a single factory

keeb:
## Summary

Fixes swamp-club lab issue **#35**.

Three production paths — manual `swamp model method run`, workflow steps, and pre-flight check validation — each hand-built a `MethodContext` literal independently. New fields added to one site silently missed the others, producing a recurring class of regression (see #947, #1145, #1148). The latest symptom: workflow steps that invoke a model calling `context.queryData` crashed with `"not a function"` because the workflow path omitted the field the manual path had just gained in #1145.

This PR:

- **Extracts `buildMethodContext`** in `src/domain/models/method_context.ts` as the single production entry point. `CommonMethodContextDeps` carries startup-scoped deps; `MethodInvocationContext` carries the per-call overlay. Field types come from `MethodContext[...]` indexed accesses so the factory adds no new infrastructure imports to the domain layer.
- **Routes all three production construction sites through the factory:**
  - `src/libswamp/models/run.ts` (manual)
  - `src/domain/workflows/execution_service.ts` — `DefaultStepExecutor.executeModelMethod` (workflow step)
  - `src/domain/models/validation_service.ts` — `DefaultModelValidationService.validateChecks` (pre-flight checks)
- **Collapses `queryData`/`dataQueryService` redundancy at the driver boundary.** `RawExecutionDriver` now derives `context.queryData` from `context.dataQueryService` with a `??` fallback, so production callers supply only `dataQueryService` while existing test fixtures that set `queryData` directly continue to work unchanged. `ModelMethodRunDeps` loses its `queryData` field; the CLI and serve deps pass `dataQueryService` instead. `CheckValidationContext` gains an optional `dataQueryService` so pre-flight checks can query data too.
- **Guards against future regressions** with `method_context_arch_test.ts`, which walks `src/` and fails on any new inline `MethodContext = {` literal outside tests, the factory, and the `MethodContext` interface declaration.

### Scope expansion

This PR deliberately bundles the bug fix, the three-site consolidation, and the queryData/dataQueryService dedup in one change, per issue-lifecycle discussion. The goal was a one-shot fix for the recurring regression class — not a follow-up trail. A separate issue (lab #84) was filed for the parallel divergence in `MethodReportContext`, which is out of scope here.

## Test plan

- [x] 5 new factory unit tests in `method_context_test.ts` — field pass-through for all required and optional `MethodContext` fields, including the invariant that the factory does not populate `queryData`.
- [x] 2 new driver tests in `raw_execution_driver_test.ts` — (1) explicit `queryData` wins over the derived binding; (2) derivation from `dataQueryService` works when `queryData` is unset, with the correct predicate and `select` pass-through.
- [x] New workflow-path unit regression in `execution_service_test.ts` — registers a capturing model, runs `DefaultStepExecutor` against a real `StepExecutionContext`, asserts the captured `MethodContext` has `dataQueryService` set and a callable `queryData`.
- [x] New integration test `integration/workflow_query_data_context_test.ts` — seeds a real `FileSystemUnifiedDataRepository` + `CatalogStore`, builds a `MethodContext` via the factory, runs an inline `MethodExecutor` through `RawExecutionDriver` against the predicate `attributes.sentinel == \"expected\"`, and verifies the seeded row comes back with attributes intact.
- [x] New architecture guard `method_context_arch_test.ts` — walks `src/`, fails on any new inline `MethodContext = {` literal outside the allowlist.
- [x] `deno check` — clean
- [x] `deno lint` — 1025 files, clean
- [x] `deno fmt --check` — 1039 files, clean
- [x] `deno run test` — **4296 passed, 0 failed** (DDD layer ratchet still at 25 — the factory reuses `MethodContext[...]` indexed types and adds no new infrastructure imports)
- [x] `deno run compile` — binary built

### DIFF
--- a/integration/workflow_query_data_context_test.ts
+++ b/integration/workflow_query_data_context_test.ts
@@ -1,0 +1,208 @@
+// Swamp, an Automation Framework
+// Copyright (C) 2026 System Initiative, Inc.
+//
+// This file is part of Swamp.
+//
+// Swamp is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License version 3
+// as published by the Free Software Foundation, with the Swamp
+// Extension and Definition Exception (found in the "COPYING-EXCEPTION"
+// file).
+//
+// Swamp is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with Swamp.  If not, see <https://www.gnu.org/licenses/>.
+
+// Integration test for the full queryData chain on the workflow path:
+// buildMethodContext -> RawExecutionDriver -> derived queryData ->
+// real DataQueryService -> real CatalogStore. A regression in any of these
+// layers would surface here. See lab issue #35.
+
+import { assertEquals, assertExists } from "@std/assert";
+import { ensureDir } from "@std/fs";
+import { join } from "@std/path";
+import { getLogger } from "@logtape/logtape";
+import { z } from "zod";
+
+import { buildMethodContext } from "../src/domain/models/method_context.ts";
+import {
+  type MethodExecutor,
+  RawExecutionDriver,
+} from "../src/domain/drivers/raw_execution_driver.ts";
+import { Definition } from "../src/domain/definitions/definition.ts";
+import { Data } from "../src/domain/data/data.ts";
+import { ModelType } from "../src/domain/models/model_type.ts";
+import type {
+  MethodContext,
+  MethodDefinition,
+  ModelDefinition,
+} from "../src/domain/models/model.ts";
+import type { DataRecord } from "../src/domain/data/data_record.ts";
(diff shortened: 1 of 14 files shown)


stack72:
Thanks for the detailed write-up @magistr — the confusion between `telemetryEndpoint` (product analytics) and OTLP tracing is a real paper cut, and we agree that's worth addressing in documentation (see #1163).

For the `.swamp.yaml` tracing block itself: we looked into this and the core challenge is that OTLP tracing initializes in `main.ts` before CLI argument parsing and before `.swamp.yaml` is loaded. Moving tracing initialization later would lose the initial startup traces, which are some of the most valuable for debugging. Reading `.swamp.yaml` earlier introduces a second marker-read path before the CLI is set up, which is a fragile change we're not comfortable making — any failure there (permissions, malformed YAML, symlinks) would block the entire CLI before it even starts.

The two-phase approach (start from env vars, reconfigure after marker load) adds significant complexity to the tracing lifecycle, and OTel providers aren't designed to be reconfigured mid-flight.

Given all of that, this isn't the right time for us to take on a rewrite of the initialization flow for what is ultimately a convenience over the standard `OTEL_*` environment variables. The env vars work well with direnv (`.envrc`) for per-repo tracing config that lives alongside the repo without changing swamp's startup path.

We're going to close this for now, but if the init flow changes for other reasons in the future, we'd revisit this idea then.

PR #1173, opened 2026-04-13
feat: swamp open — local web UI for extensions, workflows, vaults

johnrwatson:
## Summary

Adds `swamp open` — a top-level command (also registered as `swamp serve open`) that starts a localhost web UI on port 9191 for browsing and operating a swamp repository from a browser. It's a thin HTTP + SPA layer over libswamp: every operation goes through the same generators and deps the CLI already uses.

Everything renders in a themed dark UI inspired by the swamp-club `/manual` page — Orbitron + JetBrains Mono, neon green on black, cyber-border panels, scanline background.

### Features

**Repo picker** — Boots even when run outside a swamp repo. Lists existing swamps found on disk, plus any paths in optional `~/.config/swamp/index.yaml`. Per-repo metadata sidecar (`.swamp/serve_open_meta.yaml`) with name, description, tags for search.

**Extensions tab** — Sidebar of installed + registry extensions with client-side filtering. Uninstalled rows open a rich detail pane (models, methods, arguments, workflows, vaults, datastores, drivers, reports, skills, release notes) sourced from the swamp-club `/latest` endpoint when the user is logged in. Install via `pullExtension` with hot-reload of `modelRegistry` / `vaultTypeRegistry` / `driverTypeRegistry` / `reportRegistry` so new types appear without restarting. Multi-type extensions get a type picker. Model instances support create / edit (including rename that preserves the ID and history) / delete.

**Workflows tab** — Sidebar of all workflows, click one to see jobs + an inputs form. Run triggers an SSE stream rendered in a unified **task follower**: top meta strip, DAG job graph, per-job DAG of steps with topological column layering and rectilinear edges, expandable step rows with logs, errors, data artifacts, and reports. Click-to-center, scroll-preservation across live re-renders. Historical workflow runs and model-method runs both render through the same component via run-state mappers.

**Reports tab** — Lists registered reports (built-in + extension-provided) with scope and labels.

**Vaults** — Modal with a custom themed combobox (native `<select>` can't be reliably themed on macOS). Dynamic config form rendered per-provider from its Zod schema via `zodToJsonSchema` so AWS SM, 1Password, Azure KV etc. can be configured in-UI. Install additional vault providers from the registry inline. Every method/workflow input field has a 🔒 picker that fills the value with ``${{ vault.get(...) }}`` so swamp's runtime expression evaluator resolves the secret server-side.

**Auth** — Header shows the logged-in swamp-club username via `libswamp/auth.whoami`.

### Architecture notes

- All imports in `src/serve/open/http.ts` go through `src/libswamp/mod.ts` per CLAUDE.md.
- `configureExtensionLoaders` and `configureExtensionAutoResolver` extracted from `src/cli/mod.ts` so serve-open can reconfigure the global registries when the user picks a repo at runtime (the CLI bootstrap binds them to the launch cwd, which may not be the selected repo).
- Favicon SVG is embedded as a string constant so the compiled binary can serve it without filesystem access.
- UI has zero npm dependencies — single HTML file with inline CSS + vanilla JS, served as a string literal from `src/serve/open/ui.ts`. Fonts loaded from Google Fonts CDN.
- All browser `alert()` dialogs replaced with a themed toast system.

### New commands / endpoints

- `swamp open` (alias: `swamp serve open`) — starts the server
- HTTP endpoints (all `GET` unless noted): `/api/repo/{status,discover,use,init,meta}`, `/api/fs/list`, `/api/extensions/{installed,search,:name,install}`, `/api/types[?prefix]`, `/api/types/:name/describe`, `/api/definitions[?type]` (+ `POST`, `GET/:id`, `PUT/:id`, `DELETE/:id`), `/api/models/:name/{methods,methods/:m/{describe,run},history,outputs/:id}`, `/api/workflows[/:name][/describe][/run][/history][/runs/:id]`, `/api/reports`, `/api/vaults` (+ `POST`, `POST /:name/keys`), `/api/vault-types[/registry][/:type/schema]`, `/api/whoami`, `/favicon.svg`.

## Test plan

- [x] `deno lint` passes
- [x] `deno run test` — all 4305 tests pass
- [x] `deno check` passes
- [x] `deno run compile` produces a working binary
- [ ] Manual smoke test: `./swamp open` from a non-repo directory shows the picker, and from inside a repo loads the main UI
- [ ] Manual smoke test: run a model method and see the task follower render live events
- [ ] Manual smoke test: run a workflow (dag-showcase style) and see the DAG animate through state changes
- [ ] Manual smoke test: create a `local_encryption` vault, add a key, use it in a method run

### DIFF
--- a/src/cli/commands/open.ts
+++ b/src/cli/commands/open.ts
@@ -81,21 +81,11 @@ function forceExtensionCatalogRescan(repoDir: string): void {
 
 async function reloadExtensionRegistries(): Promise<void> {
   // Force the registries to re-run their loaders so newly pulled
-  // extensions are picked up without restarting the server. The private
-  // flags are reset via an unchecked cast — intentional for this POC.
-  for (
-    const reg of [
-      modelRegistry,
-      vaultTypeRegistry,
-      driverTypeRegistry,
-      reportRegistry,
-    ] as unknown as Array<
-      { extensionsLoaded: boolean; extensionLoadPromise: Promise<void> | null }
-    >
-  ) {
-    reg.extensionsLoaded = false;
-    reg.extensionLoadPromise = null;
-  }
+  // extensions are picked up without restarting the server.
+  modelRegistry.resetLoadedFlag();
+  vaultTypeRegistry.resetLoadedFlag();
+  driverTypeRegistry.resetLoadedFlag();
+  reportRegistry.resetLoadedFlag();
   await Promise.all([
     modelRegistry.ensureLoaded(),
     vaultTypeRegistry.ensureLoaded(),
--- a/src/cli/commands/open.ts
+++ b/src/cli/commands/open.ts
@@ -131,8 +131,15 @@ async function loadRepoIntoState(
 
 export const openCommand = new Command()
   .name("open")
-  .description("Start a local web UI for browsing and running extensions")
-  .example("Start and open in browser", "swamp open")
+  .description(
+    "Start a local web UI for browsing extensions, workflows, vaults, and reports",
+  )
+  .example("Open the current repo", "swamp open")
+  .example(
+    "Browse without a repo (picker mode)",
+    "cd /tmp && swamp open",
+  )
+  .example("Point at a specific repo", "swamp open --repo-dir /path/to/repo")
   .example("Custom port", "swamp open --port 9192")
   .option("--repo-dir <dir:string>", "Repository directory", { default: "." })
   .option("--port <port:number>", "Port to listen on", { default: 9191 })
@@ -282,8 +289,14 @@ export const openCommand = new Command()
     const shutdown = () => {
       if (shuttingDown) return;
       shuttingDown = true;
+      if (isJson) {
+        console.log(JSON.stringify({ status: "stopping" }));
+      }
       logger.info("Shutting down...");
       ac.abort();
+      if (isJson) {
+        console.log(JSON.stringify({ status: "stopped" }));
+      }
     };
     Deno.addSignalListener("SIGINT", shutdown);
     Deno.addSignalListener("SIGTERM", shutdown);
--- a/src/cli/commands/open.ts
+++ b/src/cli/commands/open.ts
@@ -58,7 +58,7 @@ import { VERSION } from "./version.ts";
 // deno-lint-ignore no-explicit-any
 type AnyOptions = any;
 
-const logger = getSwampLogger(["serve", "open"]);
+const logger = getSwampLogger(["open"]);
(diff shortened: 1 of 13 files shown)


johnrwatson:
Thanks for the review — quick response to each remaining item on commit 3a74e6c:

### Blocking

1. **Import boundary** — fixed in 01764cc (`zodToJsonSchema` now imported from `libswamp/mod.ts`).
2. **"Unrelated regression: Promise detection removed"** — this is a false positive. This branch has **zero** changes to `src/domain/expressions/cel_evaluator.ts`, `src/domain/workflows/execution_service.ts`, `cel_evaluator_test.ts`, or `execution_service_test.ts`. Verified with:
   ```
   $ git diff main...HEAD -- src/domain/expressions/cel_evaluator.ts src/domain/workflows/execution_service.ts
   (empty)
   $ git log main..HEAD --oneline -- src/domain/expressions src/domain/workflows/execution_service.ts
   (empty)
   ```
   No fix needed — please re-run on the latest HEAD.
3. **Outbound fetch timeout** — fixed in 01764cc with `signal: AbortSignal.timeout(10_000)`.
4. **Tests for new server code** — fixed in 01764cc (`src/serve/open/http_test.ts` covering HTML serving, favicon, repo status, 412/404/403 guards, filesystem listing, meta PUT validation) + a new `ModelRegistry.resetLoadedFlag` test in `model_test.ts`. This push (4bd9059) adds two more CORS guard tests.
5. **Private-field cast on registries** — fixed in 01764cc with public `resetLoadedFlag()` methods on `ModelRegistry`, `VaultTypeRegistry`, `DriverTypeRegistry`, and `ReportRegistry`.

### Suggestions

1. **`as any` casts in workflow/vault handlers** — fixed in 01764cc. Workflow handlers now typed with `WorkflowRun`/`JobRun`/`StepRun` from the domain layer; the vault lazy-types cast dropped (method is already public). The remaining `AnyOptions` alias is the established Cliffy workaround pattern used by every other command in the repo.
2. **CORS / Origin protection** — addressed in this push (4bd9059). `handleOpenRequest` now rejects any request whose `Origin` header doesn't match the server host with a 403.
3. **Filesystem browser scope** — left as-is for the POC. The server is localhost-only, the Origin guard blocks cross-origin exploitation, and restricting to `$HOME` would break the existing picker-mode flow where users point `swamp open` at repos on external volumes.
4. **`handleRepoMetaPut` length limits** — addressed in this push (4bd9059). Name capped at 200 chars, description at 2000, tags at 64 × 64.
5. **Route dispatch extraction** — left as a follow-up. Style preference, not a behavioral change.

PR #1211, opened 2026-04-22
feat(skills): add swamp-extension-quality scorecard guide

stack72:
## Summary

- New `swamp-extension-quality` skill that guides extension authors to hit the Swamp Club quality scorecard (11-factor rubric, third-party ceiling 12/13 = 92% Grade A).
- Scope-bounded so it doesn't overlap with `swamp-extension-{model,driver,vault,datastore,publish}` or `swamp-troubleshooting` — the SKILL.md opens with a "when this skill fires vs. when another one does" table.
- Cross-reference row added to each sibling extension skill's "When to Use Other Skills" table (`Quality scorecard & best practices → swamp-extension-quality`).

## What's inside

- `SKILL.md` — scope boundaries, factor-to-action map (11 rows), pre-publish 13-step checklist, troubleshooting for each "row shows 0/n" pattern.
- `references/rubric.md` — per-factor mechanics (exact criteria, tarball layout, host allowlist), grade thresholds, ceiling math, worked 92% example.
- `references/templates.md` — ready-to-adapt `manifest.yaml`, `README.md` (≥500 chars, ≥2 fenced blocks), JSDoc-annotated entrypoint skeleton, pre-publish command sequence, common-mistakes fix table.
- `evals/trigger_evals.json` — 20 positive triggers + 10 negative triggers (covering the overlapping-skill boundaries).

## Validation

- `tessl skill review` passed at **100%** (Description 100%, Content 100%).
- `deno fmt` clean on all modified skill markdown.

## Test plan

- [ ] Spot-check each sibling skill's "When to Use Other Skills" table renders correctly.
- [ ] Trigger the skill with one positive-eval query and confirm it fires.
- [ ] Trigger with one negative-eval query (e.g. "How do I publish my extension to the registry?") and confirm it does NOT fire (publish skill does).

### DIFF
--- a/.claude/skills/swamp-extension-datastore/SKILL.md
+++ b/.claude/skills/swamp-extension-datastore/SKILL.md
@@ -211,14 +211,15 @@ verification) before allowing a push.
 
 ## When to Use Other Skills
 
-| Need                                | Use Skill                |
-| ----------------------------------- | ------------------------ |
-| Use existing datastores             | `swamp-repo`             |
-| Create custom models                | `swamp-extension-model`  |
-| Create custom execution drivers     | `swamp-extension-driver` |
-| Repository setup and configuration  | `swamp-repo`             |
-| Manage secrets for datastore config | `swamp-vault`            |
-| Understand swamp internals          | `swamp-troubleshooting`  |
+| Need                                | Use Skill                 |
+| ----------------------------------- | ------------------------- |
+| Use existing datastores             | `swamp-repo`              |
+| Create custom models                | `swamp-extension-model`   |
+| Create custom execution drivers     | `swamp-extension-driver`  |
+| Repository setup and configuration  | `swamp-repo`              |
+| Manage secrets for datastore config | `swamp-vault`             |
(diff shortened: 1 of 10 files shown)


stack72:
## CI fix appended: `f511238c`

Main CI is currently broken on `integration/extension_rm_test.ts` for **every open PR**, not just this one. #1210 hits the same failure with no code overlap.

**Root cause** (not introduced by this PR): `pull.ts:992-1016` intentionally records each pulled skill as a single directory-root path in `upstream_extensions.json` so `extension rm` can delete the skill in one recursive `Deno.remove` (rm.ts:337-339 already handles `stat.isDirectory`). The test predated that design and still asserts `stat.isFile === true`. The moment `@keeb/ssh` was republished with a directory-shaped skill, the assertion started failing.

**Fix** (2-line test change): accept `stat.isFile || stat.isDirectory` before rm. Post-rm assertion already tolerates both (relies on `Deno.stat` throwing on absence).

I bundled the fix here rather than splitting into a separate PR because (a) every open PR is currently blocked on it so it needs to land ASAP, and (b) the commit is cleanly separated so the history shows intent.

PR #1216, opened 2026-04-23
fix(datastore): bound extension sync calls with AbortSignal timeout (swamp-club#157)

stack72:
## Summary

Gives swamp core a hard deadline on every extension datastore sync call (push and pull). Prevents the ~8.5-minute hang + upstream Deno TLS panic reported in **swamp-club#157** by bounding time independent of whether the extension honors the AbortSignal.

Fixes **swamp-club#157**.

## Design

**Single-AbortController pattern** — one `setTimeout` drives `controller.abort(new SyncTimeoutError(...))`. Two effects from one timer:

- Signal-compliant extensions receive a real `AbortSignal` and can unwind cooperatively.
- Non-compliant extensions are bounded by `Promise.race` listening on `controller.signal.reason`, so the deadline fires regardless of extension cooperation.

One source of truth for the deadline. No ms-level drift between independent timers.

**Covers every entry point** that invokes an extension sync method:

| Path | File | Behavior on timeout |
|---|---|---|
| End-of-command push (flush) | `src/infrastructure/persistence/datastore_sync_coordinator.ts:flushDatastoreSyncNamed` | `SyncTimeoutError` propagates, exit 1 |
| Start-of-command pull (register) | `src/infrastructure/persistence/datastore_sync_coordinator.ts:registerDatastoreSyncNamed` | `SyncTimeoutError` propagates, exit 1 |
| Setup migration push | `src/libswamp/datastores/setup.ts` | Collected into setup's existing warnings bucket, exit 0 |
| `swamp datastore sync --push/--pull` | `src/libswamp/datastores/sync.ts` | `SyncTimeoutError` propagates, exit 1 |

Asymmetric error surface is deliberate: **timeouts always propagate**, but other push errors in the coordinator flush path still warn-downgrade (transient S3 blips should not break a successful command).

**Default timeout**: 300s (5 minutes), configurable via `SWAMP_DATASTORE_SYNC_TIMEOUT_MS` env var or per-datastore `syncTimeoutMs` field on `CustomDatastoreConfig`. This is provisional — intended to be tuned down once production telemetry grounds the slow-path distribution.

## Behavior Changes (heads-up for release notes)

1. **Timeouts now fail loudly.** Commands whose push silently hangs today will exit 1 with `SyncTimeoutError` after the timeout fires. Teams whose CI was tolerating silent push failures will see red. This is the intended direction — silent data loss is worse than visible failure.
2. **`swamp datastore sync` exits 1 on hang** (was: hang indefinitely).
3. **`swamp datastore setup extension ...` completes-with-warning on push hang** (was: hang indefinitely). Local cache migrates successfully; user re-runs `swamp datastore sync --push` after clearing the underlying issue.

### Unit + integration tests (4720 passed, 0 failed, 1 ignored)

New tests in `src/infrastructure/persistence/datastore_sync_coordinator_test.ts` cover:

- Hanging push surfaces `SyncTimeoutError` within configured window
- Error message contains \`swamp datastore lock release --force\` hint
- Signal-compliant extension: abort propagates, outer race still wins with `SyncTimeoutError` (not extension's `AbortError`)
- Lock released on timeout path — next run is not stranded
- Hanging pull surfaces `SyncTimeoutError` through pull summary wrapper
- Normal settlement does not leak timers (verified by Deno's leak detector)
- `resolveSyncTimeoutMs`: config field > env > default precedence
- Invalid env values fall back to default
- End-to-end config→timer integration test
- Multi-entry flush: one entry's timeout does not strand other entries' locks

### End-to-end synthetic reproduction against compiled binary

Built a minimal \`@test/hang-datastore\` extension whose \`pushChanged\` returns \`new Promise(() => {})\` (never resolves). Ran three commands against the locally-compiled \`./swamp\` binary:

1. **\`swamp extension install\`** (the exact user scenario from swamp-club#157):
   - Bounded at configured \`SWAMP_DATASTORE_SYNC_TIMEOUT_MS=2000\`, fired on schedule
   - Exit code: **1**
   - FTL-level \`SyncTimeoutError\` with actionable hint surfaced: \`Datastore push to \"@test/hang-datastore\" timed out after 2000ms. If a prior process crashed without releasing the lock, run 'swamp datastore lock release --force' (add --model <type>/<id> for a specific model lock).\`

2. **\`swamp datastore setup extension @test/hang-datastore --config '{}'\`**:
   - Bounded at 2000ms, fired on schedule
   - Exit code: **0** (setup's pre-existing warn-semantics preserved)
   - Surfaced as \`Warnings: @test/hang-datastore push failed: Datastore push ... timed out after 2000ms\` in the setup output

3. **\`swamp datastore sync --push\`**:
   - Bounded at 2000ms, fired on schedule
   - Exit code: **1**
   - FTL-level \`SyncTimeoutError\` with actionable hint

### Real AWS verification (happy path)

Ran against a real S3 bucket in \`us-east-1\` with the production \`@swamp/s3-datastore\` extension:

| Command | Outcome | Timing |
|---|---|---|
| \`swamp datastore setup extension @swamp/s3-datastore --config '{real bucket}'\` | exit 0 | ~10s (includes extension install) |
| \`swamp extension install\` | exit 0, full pull+push through coordinator | pull 560ms, push 140ms |
| \`swamp datastore sync --pull\` | exit 0, Pulled 0 files | <1s |
| \`swamp datastore sync --push\` | exit 0, push complete | 140ms |
| \`swamp model create command/shell hello\` | exit 0, subsequent flush push succeeds | 140ms |

**Confirmed**: the timeout wrapping does not interfere with normal real-AWS S3 operations. The coordinator logs \`Pushing changes to @swamp/s3-datastore...\` exactly once per command exit, proving \`runBoundedSync\` is being invoked in the real CLI stack against the real AWS SDK middleware.

Test bucket was created fresh, populated during the run, and torn down at the end. No AWS artifacts remain.

### UAT compatibility audit (\`systeminit/swamp-uat\`)

Thorough audit of \`~/code/systeminit/swamp-uat/\` concluded **SAFE** across the suite:

- No UAT tests mock \`DatastoreSyncService\` directly (all testing is CLI-integration)
- No existing tests reference \`SyncTimeoutError\` or \`SWAMP_DATASTORE_SYNC_TIMEOUT_MS\`
- No tests assert hang durations
- Exit-code contracts hold — affected-area tests (\`tests/cli/datastore/{lock,setup,sync,status}_test.ts\`, \`tests/cli/extension/install_test.ts\`) use fast backends that don't hit timeouts
- Adversarial suite (\`concurrency_test.ts\`, \`state_corruption_test.ts\`) is orthogonal to datastore sync

A follow-up UAT issue will add coverage for \"bounded timeout under hanging push\" — additive, not a fix to an existing test.

### Checks passed

- [x] \`deno check main.ts\`
- [x] \`deno lint\` (1080 files)
- [x] \`deno fmt\` (1198 files)
- [x] \`deno run test\` (4720 passed)
- [x] \`deno run compile\` (binary at \`./swamp\`, 243MB)

## Out-of-Scope Unbounded Sites (Extension-Side)

Real AWS verification surfaced two unbounded call sites that are **not covered by this PR** because they live inside extension code rather than swamp core. Documenting them here so reviewers have the full picture:

1. **\`S3Lock.acquire()\`** — calls \`putObjectConditional\` against S3 with no per-request timeout. Hangs indefinitely on a slow/unreachable endpoint, before the coordinator's bounded pull/push code can run.
2. **\`S3DatastoreVerifier.verify()\`** — calls \`headBucket\` against S3 with no per-request timeout. Invoked by \`swamp datastore setup extension ...\` before migration, so a broken endpoint hangs setup before the bounded push path is reached.

Both sites are inside \`@swamp/s3-datastore\` and are addressed by the companion extension PR's DEF-1 (per-request \`AbortSignal.timeout\` on every S3 SDK call). Swamp core cannot bound these without introducing a cross-cutting signal contract on \`DistributedLock\` and \`DatastoreVerifier\`, which would be a larger surface change than swamp-club#157 warrants.

This PR's scope remains the sync-service contract (\`pushChanged\` / \`pullChanged\`) — the specific hang site reported in swamp-club#157.

## Known Risks

| Risk | Severity | Status |
|---|---|---|
| Teams silently tolerating failed pushes see new exit 1 on coordinator/sync paths | Medium, behavior change | Documented in release notes |
| Setup migration completes-with-warning on hang instead of hanging — user ends up with successful local cache but incomplete remote push | Medium, behavior change | Documented; recovery is \`swamp datastore sync --push\` |
| 300s default is a guess — no production telemetry yet | Low | Documented as provisional |
| Setup uses env-or-default for timeout, not per-config \`syncTimeoutMs\` (config isn't fully materialized during setup) | Low | Minor inconsistency |
| Lock release on timeout can itself hang if extension's \`release()\` is broken | Low | Depends on extension fix (separate PR) |
| Lock **acquire** + datastore **verify** paths hang on slow endpoints (extension-side) | Low | Out of scope — fixed by companion extension PR's DEF-1 |

## Companion Work

- **\`@swamp/s3-datastore\` extension fix** (separate repo, different agent): adds per-request timeouts on S3 SDK calls (covers lock acquire/release and verify paths), retry budget on \`pushFile\`, stale-lock steal backoff with conditional delete, conditional lock release. That's the root-cause fix; this PR is the defense-in-depth layer for the sync-service contract specifically.
- **UAT coverage** (follow-up issue in \`systeminit/swamp-uat\`): bounded-timeout-under-hanging-push test.
- **Upstream Deno \`tls_wrap.rs:1918\` unwrap bug**: our bounding window avoids reaching the retry depth that exposes the panic. If upstream fix exists, follow-up PR to bump the pinned Deno version.

## Test plan

- [x] Unit tests: 18 new scenarios covering the bounded-sync mechanism, error propagation, and config resolution
- [x] Integration: full \`deno run test\` (4720 pass)
- [x] End-to-end synthetic: hang extension against compiled binary, three distinct command paths
- [x] End-to-end real AWS: happy path against real S3 bucket (created, exercised, torn down)
- [x] UAT compatibility: audit shows no regressions
- [ ] Post-merge: file UAT gap issue in \`systeminit/swamp-uat\`

### DIFF
--- a/.claude/skills/swamp-repo/references/troubleshooting.md
+++ b/.claude/skills/swamp-repo/references/troubleshooting.md
@@ -6,6 +6,7 @@
   - ["Not a swamp repository"](#not-a-swamp-repository)
   - [Config File Issues](#config-file-issues)
   - [Skills Not Loading](#skills-not-loading)
+  - [Stuck Datastore Lock](#stuck-datastore-lock)
 - [Recovery Procedures](#recovery-procedures)
 
 ## Common Errors
@@ -99,6 +100,54 @@ ls -la .claude/skills/
    ```bash
    swamp repo upgrade --json
    ```
+
+### Stuck Datastore Lock
+
+**Symptom**: `swamp` commands hang on `datastore·sync Pushing changes to ...`,
+or fail with `SyncTimeoutError: Datastore push/pull timed out after Nms`, or
+livelock on `Global lock held by <holder> appears stale ... — proceeding`
+followed by repeated attempts to acquire.
+
+**Cause**: A previous `swamp` process crashed or was killed before releasing the
+datastore lock. The lock file/object in S3 or the filesystem still names a
+holder that is no longer running.
+
+**Diagnose**:
+
+```bash
+swamp datastore lock status
+```
+
+Shows the holder (hostname, pid, acquiredAt, ttlMs). If the holder is on a
+host/runner that is no longer running, the lock is orphaned.
+
+**Solutions**:
+
+1. **Wait for TTL expiry**: Locks self-expire after `ttlMs` (30s default).
+   Non-interactive — just wait.
+
+2. **Force-release the global lock** (breakglass):
+
+   ```bash
+   swamp datastore lock release --force
+   ```
+
+   The `--force` flag is required. Bypasses `release()` and deletes the lock
+   directly. Use when a crashed process left a lock that hasn't expired yet.
+
+3. **Force-release a specific per-model lock**:
+
+   ```bash
+   swamp datastore lock release --force --model <type>/<id>
+   ```
+
+   Example: `swamp datastore lock release --force --model aws-ec2/my-server`.
+
+**Related knobs**:
+
+- `SWAMP_DATASTORE_SYNC_TIMEOUT_MS` — hard deadline per direction (push or pull)
+  in milliseconds. Default 300000 (5 minutes). Increase for very large
+  datastores on slow networks; decrease to fail faster in tight CI loops.
 
 ## Recovery Procedures
 
--- a/design/datastores.md
+++ b/design/datastores.md
@@ -168,6 +168,25 @@ datastore:
     - "telemetry/**"
 ```
 
+### Sync Timeout
+
+Each direction of a remote sync (push / pull) is bounded by a hard deadline
+enforced in the coordinator, so a stuck or slow extension cannot hang the CLI
+indefinitely. The default is `DEFAULT_SYNC_TIMEOUT_MS` (5 minutes), overridable
+per-datastore on `CustomDatastoreConfig.syncTimeoutMs` or globally via the
+`SWAMP_DATASTORE_SYNC_TIMEOUT_MS` environment variable.
+
+The deadline fires a `SyncTimeoutError` regardless of whether the extension
+honored the `AbortSignal` passed to `pushChanged(options)` / `pullChanged(options)`.
+Timeouts propagate as a non-zero CLI exit so the user sees that data did not
+make it to the remote; other push errors still warn-downgrade (preserves
+historical behavior where a transient S3 blip does not kill a run).
+
+See `src/domain/datastore/datastore_config.ts` (`DEFAULT_SYNC_TIMEOUT_MS`,
+`SYNC_TIMEOUT_ENV_VAR`, `resolveSyncTimeoutMs`) and
+`src/infrastructure/persistence/datastore_sync_coordinator.ts`
+(`runBoundedSync`).
+
 ## Path Resolution
 
 Every file operation goes through a `DatastorePathResolver` that decides whether
--- a/packages/testing/datastore_types.ts
+++ b/packages/testing/datastore_types.ts
@@ -66,10 +66,15 @@ export interface DatastoreVerifier {
   verify(): Promise<DatastoreHealthResult>;
 }
 
+/** Options accepted by sync service methods. */
+export interface DatastoreSyncOptions {
+  signal?: AbortSignal;
+}
+
 /** Interface for datastore synchronization services. */
 export interface DatastoreSyncService {
-  pullChanged(): Promise<number | void>;
-  pushChanged(): Promise<number | void>;
+  pullChanged(options?: DatastoreSyncOptions): Promise<number | void>;
+  pushChanged(options?: DatastoreSyncOptions): Promise<number | void>;
 }
 
 /**
--- a/src/cli/mod.ts
+++ b/src/cli/mod.ts
@@ -1062,8 +1062,17 @@ export async function runCli(args: string[]): Promise<void> {
       }
     }
   } catch (error) {
-    // Release datastore lock even on failure (don't leave locks stuck)
-    await flushDatastoreSync();
+    // Release datastore lock even on failure (don't leave locks stuck).
+    // flushDatastoreSync() can now throw (SyncTimeoutError propagates on
+    // the push path — see swamp#1216), so we swallow errors here: the
+    // original command error must take precedence, and telemetry
+    // recording below must still run. Per-entry cleanup on
+    // flushDatastoreSync failure is handled inside the coordinator.
+    try {
+      await flushDatastoreSync();
+    } catch {
+      // Best effort — don't shadow the original error.
+    }
 
     // Record error invocation and flush before re-throwing
     if (telemetryCtx && error instanceof Error) {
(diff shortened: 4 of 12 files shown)


stack72:
Addressed the double-render + stack-trace review feedback in 6a83ad26.

**Three changes:**

1. `SyncTimeoutError` now extends `UserError` (was: `Error`) — `renderError` routes it through the message-only branch, no stack trace.
2. Coordinator flush path: removed the redundant `logger.error(...)` for `SyncTimeoutError`; the top-level renderer produces the single line. Warn-downgrade for non-timeout push errors is preserved.
3. Coordinator pull path: `SyncTimeoutError` now rethrows directly instead of being wrapped in a generic `Error` (the wrap would've lost the clean render).

**Plus** a regression test asserting `SyncTimeoutError instanceof UserError` so a future "fix" back to `extends Error` gets caught at CI.

**Verified against the compiled binary** — `swamp extension install` against a hanging datastore now renders exactly one line, exit 1:

```
FTL error Error: `Datastore push to "@test/hang-datastore" timed out after 2000ms. If a prior process crashed without releasing the lock, run 'swamp datastore lock release --force' (add --model <type>/<id> for a specific model lock).`
```

No stack trace, escape hatch front-and-center, no coordinator summary line doubling it. `deno run test` still green (4721 passed).

stack72:
Addressed the orphaned-entry / double-timeout finding in 7b92439c. Both fixes applied:

**Option A (root cause)** — `registerDatastoreSyncNamed` now has an `unwindOnFailure` helper that removes the entry from the map, releases the lock if it was acquired, and deinstalls the signal handler if no entries remain. Covers both failure paths (lock-acquire failure and pull failure) so the coordinator can never leak an entry.

**Option B (safety net)** — `runCli()`'s error-handler `flushDatastoreSync()` call is now wrapped in `try/catch`. Even if some future code path violates the entry-cleanup invariant again, the original command error takes precedence and telemetry still runs.

**Regression tests** added:
- `registerDatastoreSyncNamed: pull timeout cleans up entry and releases lock` — asserts the entry is gone from `getRegisteredLockKeys()` after the throw, lock was released exactly once, and a subsequent `flushDatastoreSync()` is a no-op.
- `registerDatastoreSyncNamed: lock acquire failure removes entry` — symmetric check for the pre-lock failure path.

**End-to-end verified** against the compiled binary with a synthetic `hang-pull-datastore` extension (pull hangs, push is fine):

```
exit: 1  elapsed: 3s
23:40:52.219 INF datastore·sync Syncing from "@test/hang-pull-datastore"...
23:40:54.636 FTL error Error: `Datastore pull to "@test/hang-pull-datastore" timed out after 2000ms. If a prior process crashed without releasing the lock, run 'swamp datastore lock release --force' ...
```

With `SYNC_TIMEOUT_MS=2000`, total elapsed is ~3s (pull timeout + process overhead). Before this fix, the error handler would have hit push and waited another 2s = 4s (or in production with 300s defaults: 10 minutes). Single error, no shadowing, no double-wait.

`deno run test` still green (4723 passed).

PR #1239, opened 2026-04-29
agent-efficiency: 12 commits, ~33% turn reduction on the k8s-debug-v2 bench

johnrwatson:
## TL;DR

Twelve commits aimed at reducing the turn-count, cost, and wall-time
that LLM agents (specifically headless `claude`) spend driving swamp.
Every change is backed by transcript-level evidence from the
[swamp-benchmark](https://github.com/systeminit/swamp-benchmark)
`k8s-debug-v2` challenge — a deliberately-broken GKE namespace that
the agent must repair using only swamp.

**Bench trajectory (cold-start, 5 runs each, mean):**

| binary state                                        | turns | cost   | wall   | pass |
| --------------------------------------------------- | ----: | -----: | -----: | ---- |
| pre-PR baseline (April 2026, no changes)            |   145 | $2.15  | 8m37s  | 1/1  |
| infra commits only (waves 1+2, no docs)             |   121 | $1.41  | 6m08s  | 3/3  |
| infra + cheat-sheet docs                            |   119 | $1.50  | 7m03s  | 5/5  |
| **infra + docs + Autopilot fix in bench framework** | **97** | **$1.34** | **5m22s** | **5/5** |

**Net effect of this PR (final batch vs baseline):** **−33% turns,
−38% cost, −38% wall.** Best run in the final batch was 83 turns /
$0.96 / 4m00s.

The Autopilot fix lives in the bench framework, not here — but the
infra+docs work in this PR is what the agents needed to actually
*use* the cleaner state.

---

## What's in this PR

Twelve commits, grouped by theme. Each headline includes the bench
evidence I used to size the change.

#### `0cd9cfe` feat(cli): suggest semantic option aliases on unknown flags

Cliffy's default "Did you mean" suggestion is purely lexical, so
typing `--arg` on `swamp model method run` got pointed at `--log`
(3 chars from `--arg`) instead of `--input` (5 chars). The bench
showed agents losing ~3 turns to this guess-the-flag dance,
because three commands spell the same concept differently
(`--global-arg` on `model create`, `--input` on `model method run`
and `workflow run`).

Adds a small alias map (`--arg`/`--args`/`--inputs`/`--param[s]`
→ `--input`/`--global-arg`), filtered to flags actually defined
on the current command, preferred over the lexical match. Falls
back to the existing edit-distance suggestion when no semantic
alias applies.

```
$ swamp model method run X foo --arg key=val
error: Unknown option "--arg". Did you mean "--input"?
```

Wired into the existing `unknownCommandErrorHandler` (already
attached to `model`, `method`, `workflow`, `vault`, `extension`,
`report` command trees). 9 unit tests.

#### `6515b60` feat(types): add `--methods` flag for compact method listing

When an agent encounters an unfamiliar model type, the first
question is "what methods does this have?". The full
`swamp model type describe X --json` is ~190 KB for a
13-method extension; agents grep/jq/python-piped it three
different ways before extracting names. Bench transcript shows
9 calls per run on `--json`-then-pipe.

Adds `--methods`:
```
$ swamp model type describe @john/namespace --methods
list                — List all namespaces
get                 — Get one namespace
health              — Aggregated namespace health
…
```

In `--json` mode the output is `{ type, version, methods: [{
name, description }] }` — for `@john/namespace` that's ~2 KB
vs ~190 KB for the full form (90× smaller). Adds 3 unit tests.

#### `5744f47` feat(types): add `--method <name>` for single-method describe

Follow-up to `--methods`. Once an agent picks a method, it
needs the argument schema for that one method — and today's
only path is to re-pipe the full `--json` describe. The April
bench shows 7-13 such re-pipes per run.

Adds `--method <name>`:
```
$ swamp model type describe @john/namespace --method health --json
{ "type": ..., "version": ...,
  "method": { name, description, arguments: { JSON Schema },
              dataOutputSpecs: [...] } }
```

For `@john/namespace.health` the output is ~13.5 KB vs ~190 KB
for the full describe (~14× smaller). Mutually exclusive with
`--methods` via cliffy `conflicts`. Unknown method names throw
a UserError listing available names. 2 unit tests.

#### `0c2de8d` feat(data): add `--content` flag to `swamp data list`

`swamp data list <model>` returns metadata only; to see what's
in each record, callers fan out to one `swamp data get` per
item. The April bench shows up to 23 sequential `data get`
calls per run for this exact pattern.

Adds `--content`:
```
$ swamp data list my-ns-pod --content --json
{ groups: [ { type: resource,
              items: [
                { name: pod-a, version: 2,
                  content: { phase: "Running", … }, … },
                …
              ] } ] }
```

JSON content is parsed; everything else returned as a UTF-8
string. A `--max-content-bytes` knob (default 1 MiB) skips
oversized items so the flag is safe on logs/files. The fetch
path reuses `UnifiedDataRepository.getContent`. Adds the
`getContent` dependency on `DataListDeps` (optional, so
existing callers/tests don't break) and 4 unit tests.

#### `ad53bb2` feat(workflow): summarize data produced after a run

After `swamp workflow run`, callers had no way to see what the
run wrote without fanning out across `data list` / `data get`.
The April bench shows agents spending 10-15 turns just
re-discovering data the workflow had just produced.

Adds a `dataProduced` aggregate to the run output:
* JSON: top-level `dataProduced: { totalCount, entries: [{
  modelName, specName, count }] }`.
* Log mode: a compact table at the end, replacing the previous
  one-line-per-artifact dump (which scaled badly: 28 unique
  names → 28 lines of `swamp data get …`).

```
Wrote 28 data records across 4 model/spec pairs:
  v3-test-1-deployment.deployment   (4)
  v3-test-1-event.event            (17)
  v3-test-1-pod.pod                 (4)
  v3-test-1-service.service         (3)

Inspect with:
  swamp data list --workflow debug-namespace-deep
  swamp data get --workflow debug-namespace-deep <recordName>
```

Computed from each step's `dataArtifacts[].tags.{modelName,
specName}` so no plumbing changes in execution. 4 unit tests.

#### `598afd9` feat(workflow): actionable validate error messages

The April bench shows agents burning 1-8 validate cycles per
fix workflow; the slow runs lose 20+ turns to the
validate-fix-rerun loop. The dominant failure modes weren't
subtle CEL errors — they were the basic "workflow not found"
/ "missing required input" errors giving no guidance on how
to recover.

Three classes of validate-time errors made actionable:

1. **`workflow validate <name>` when the workflow doesn't
   exist.** Today: `"Workflow not found"`. New behaviour:
   * If the input looks like a file path (ends in `.yaml`/
     `.yml` or has a slash), explain that validate takes a
     name/ID and point at `swamp workflow create <name>`.
   * If it looks like a name and similar names exist, suggest
     the closest match and list a sample of existing workflows.
   * If no workflows exist at all, print the literal command
     to create the missing workflow.

2. **`Method 'X' not found on model type 'Y'`.** The resolver
   now passes `availableMethods` alongside `modelType`, and
   the failure message lists them. Same lesson as the
   `--methods`/`--method` flags — surface the answer at the
   point of failure.

3. **`Missing required inputs: a, b`.** The resolver now
   passes per-arg types from the JSON Schema, and the failure
   message includes them ("Missing required inputs:
   environment (string), version (string)"). Points at
   `swamp model type describe X --method Y` for the full
   schema.

Live behaviour:
```
$ swamp workflow validate workflows/fix.yaml
error: 'workflows/fix.yaml' looks like a file path. `swamp
workflow validate` takes a workflow name or ID, not a file. …

$ swamp workflow validate deploy-aplp
error: Workflow 'deploy-aplp' not found. Did you mean
'deploy-app'? Existing workflows: deploy-app. Run `swamp
workflow search` for the full list.
```

`MethodResolution.argTypes` and `MethodResolution.availableMethods`
are added as optional fields so existing callers/tests don't
have to change. Adds 3 tests for the libswamp not-found
branches and 2 for the validation-service error enrichment.

This is **phase 1** of a larger improvement (workflow template
scaffolds = phase 2). Phase 2 filed at swamp.club/lab/188.

#### `95ebaf0` perf(cli): detach telemetry flush

Tracing CLI startup: `swamp --version` is 0.23 s, but
`swamp model search --json` in an initialized repo is 0.55 s.
The entire 0.32 s gap is one synchronous HTTP POST to
telemetry.swamp.club, awaited at `mod.ts:1026` before process
exit. The code's own comment already noted local persistence
was durable and unflushed entries would retry on the next
invocation — so the await was gratuitous.

This commit:
* Adds a `swamp telemetry flush` subcommand that reuses
  `initTelemetryService` + `flushTelemetry` to ship pending
  entries on its own time.
* In `mod.ts` (success and error paths), replaces the awaited
  flush with a detached `Deno.Command` spawn of `swamp
  telemetry flush --repo-dir <dir>`, stdio nulled,
  `SWAMP_NO_TELEMETRY=1` and `SWAMP_NO_UPDATE_CHECK=1` on the
  child env to prevent fork chains. Parent exits immediately;
  orphaned child finishes the HTTP in the background.

Measured impact:
```
before: model search in repo = 0.55 s
after:  model search in repo = 0.24 s   (matches --version floor)
```

~320 ms saved per CLI call. Across the 145-call k8s-debug-v2
bench that's ~46 s of wall time per run.

### Skill/docs updates — teach the new patterns

The CLI changes above don't help agents who don't know about
them, and agents don't read `--help`. This PR includes five
docs commits that surface the new flags in the right places
in the CLAUDE.md/AGENTS.md template (which `swamp init` writes
to every new repo) and in the relevant skill files.

#### `0956d80` docs(skills): initial cheat-sheet bullets

Adds a "Hot-path commands" section to the CLAUDE.md template:
six bullets covering `--methods`, `data list --content`, the
`dataProduced` block, aggregator-method preference, the
"capture-once-parse-repeatedly" pattern, and the
`--input`/`--global-arg` flag-name table. Plus updates to the
`swamp-data`, `swamp-model`, and `swamp-workflow` skill files
with deeper guidance on the same patterns.

#### `c710ced` docs(skills): add three more cheat-sheet bullets from bench audit

After the first batch's transcript audit, three more turn-burn
patterns surfaced. Added bullets on:
* Searching for existing workflows before writing one.
* Sharper aggregator naming hints (`health`/`summary`/`*All`/
  `diagnose*`).
* Stop-and-check-for-batch-variant before fanning out method
  calls.

Cheat sheet length grew 6 → 9 bullets.

#### `eeccad6` docs(skills): wording tweaks after the wave-2 commits

Two corrections:
* Methods-discovery bullet now mentions `--method <name>`
  (shipped after the original docs commit).
* Aggregator-methods bullet previously used `getRolloutStatusAll`
  as an example, but no such method actually exists on
  `@john/k8s` — risk of agents searching for a phantom method.
  Softened to generic naming hints only.

#### `ea9bbfe` docs(skills): make swamp-getting-started intent-triggered, not state-triggered

Bench transcripts showed agents being routed into the
onboarding tutorial even when the user message clearly stated
a task ("debug the GKE cluster in namespace X using the
@john/k8s extension"). The previous wording forced
onboarding whenever `swamp model search` returned an empty
list:

> If no models are returned (empty result), you MUST
> immediately invoke the swamp-getting-started skill before
> doing anything else.

That conflates *repo state* with *user intent*. An empty model
list is normal for fresh `swamp init` repos, on-demand
extension pulls, and any automated/CI/agent context where the
task is concrete.

Rewrite the rule so onboarding triggers on user intent ("I'm
new to swamp", "walk me through it") — not repo emptiness.
Default for a concrete task is: do the task.

#### `4a562dd` docs(skills): make data list --content the explicit default

Adoption analysis on the third bench batch showed asymmetric
flag uptake — agents picked up `--method <name>` (used 1-8×
per run) but skipped `--content` (used 0-2× in slow runs vs
4-6× in fast runs). The slow runs went straight to N
sequential `data get <model> <name>` calls — the exact
pattern the flag was built to replace.

The previous bullet ("Inspect model data in one call …
Skip the data list → N×data get fan-out") was too soft.
Reword to make `data list --content` the explicit first move
and position `data get` as the narrow case (you already know
the exact record name AND don't want anything else). Calls
out N-sequential `data get` as "almost always wrong".

---

## What's NOT in this PR

The agent-efficiency arc spans three repos. This PR is one
of three:

* **`@john/k8s` extension** (johnrwatson/k8s-swamp-extension)
  — added the `health` aggregator method on `@john/namespace`
  (replaces fanning out across pods/deployments/services/events
  during verification). Released as v2026.04.29.5. Lives in
  the extension repo because that's where the model
  definitions live.

* **`swamp-benchmark` repo** (systeminit/swamp-benchmark)
  — explicit `resources.requests` on the seed deployments to
  stop GKE Autopilot from firing
  `autopilot-default-resources-mutator` Warning events on
  every spec change. Without this, runs stalled indefinitely
  with all the candidate's fixes correctly applied but the
  assertion never clearing.

Both are already merged on their respective `main` branches
and were measured as part of the bench trajectory above.

---

## Filed as lab issues

Bench-evidence-backed follow-ups that need bigger surface than
this PR:

* **swamp.club/lab/188** — phase 2 of the workflow-validate
  improvements: template scaffolds (`swamp workflow create
  --from-template <kind>`).
* **swamp.club/lab/189** — `swamp model method run --foreach`
  to fan out a method across every instance of a model. The
  per-instance fan-out pattern is the second-biggest remaining
  burn in the bench (11-22 calls per run on the wave-1 batch,
  reduced to 1-5 on the latest batch but still meaningful).

---

## Test plan

* [ ] `deno test` — new unit tests in
      `unknown_command_handler_test.ts`, `workflow_run_test.ts`,
      `data/list_test.ts`, `type_describe_test.ts`,
      `validate_test.ts`, `validation_service_test.ts`. All
      pass locally.
* [ ] Live smoke tests captured inline in commit messages
      (each commit includes the actual CLI output the agent
      now sees).
* [ ] Bench evidence: 5/5 pass on `k8s-debug-v2` cold-start
      with the final binary.


keeb (commented):
Consider modeling `docker inspect`

i.e, 

```
docker inspect -f '{{json .NetworkSettings}}' <container_id>
```

Issue #247, opened 2026-02-10
swamp type get should be replaced by swamp model type get


stack72:
This keeps consistency with #246

PR #1317, opened 2026-05-05
fix(shell): honor AbortSignal so --timeout cancels in practice (swamp-club#247)

stack72:
## Summary

Closes swamp-club#247.

PR #1305 added `--timeout` to `model method run` and `workflow run` but
documented it as cooperative-only because built-in models did not honor
the abort signal. This wires it through in two layers so the flag works
in practice and surfaces the `cancelled` envelope #1305 originally
promised.

### Layer 1 — built-in honoring AbortSignal

- `command/shell` passes `context.signal` into `executeProcess` and
  re-throws `AbortError` ahead of its silent-swallow paths so the abort
  is not buried as `exitCode = -1`.
- `executeProcess` uses `Deno.Command`'s native `signal` option (Deno
  2.7.14 sends SIGTERM, matching the existing timeout-path semantics)
  and explicitly normalizes signal-driven kills to `AbortError` on all
  three execution branches (streaming, buffered+timeoutMs, simple
  buffered). The simple-buffered branch switches from `command.output()`
  to `command.spawn() + process.output()` because the former silently
  ignores `CommandOptions.signal` under Deno 2.7.14.

### Layer 2 — driver-layer error-type preservation

Without this, the abort works but the JSON envelope reads
`code: "method_execution_failed"` instead of `cancelled`, because the
driver wire boundary flattens exceptions to a `string`.

- `ExecutionResult` gains a `cancelled?: boolean` flag.
- `RawExecutionDriver` sets it when the failure was an `AbortError`.
- `DefaultMethodExecutionService` re-throws as
  `DOMException(\"AbortError\")` when `driverResult.cancelled` is set
  (both call sites) so libswamp's `run.ts` handler routes it to
  `cancelled()`.

End-to-end output is now:

\`\`\`json
{
  \"error\": \"Operation was cancelled.\",
  \"code\": \"cancelled\"
}
\`\`\`

### Help text

Dropped the \"Cooperative — only honored by methods that check
AbortSignal\" caveat on `--timeout` for both `model method run` and
`workflow run`. Built-ins now honor it transparently; extension-model
methods still must check `ctx.signal`.

## Test Plan

- [x] 3 new `shell_model_test.ts` tests: AbortError surfaces, no data
      record on abort, `ignoreExitCode` does not swallow AbortError
- [x] 2 new `process_executor_test.ts` tests for the buffered branches
      (the streaming-mode signal test was already there)
- [x] 2 new `raw_execution_driver_test.ts` tests for the
      `cancelled`-flag contract
- [x] New `integration/shell_timeout_test.ts` drives the CLI binary
      against `sleep 30` with `--timeout 1s` and asserts
      `code: \"cancelled\"`
- [x] \`deno check\` / \`deno lint\` / \`deno fmt --check\` clean
- [x] Full suite: 5468 passed, 0 failed, 28 ignored
- [x] Manual scratch-repo: \`swamp model method run sleep10 execute
      --input run='sleep 10' --timeout 1s --json\` aborts in ~1s
      (was 10.5s) with the cancelled envelope above

## Follow-up

systeminit/swamp-uat's
\`tests/cli/adversarial/method_run_timeout_test.ts\` works around this
bug today by using a custom \`abort-aware-sleep\` extension fixture
instead of \`command/shell\`. After this PR lands and a swamp release is
tagged, that test should be updated to also exercise \`command/shell\`
directly. The existing fixture-based variant should stay as an
extension-model signal-handling regression guard.

### DIFF
--- a/integration/shell_timeout_test.ts
+++ b/integration/shell_timeout_test.ts
@@ -1,0 +1,127 @@
+// Swamp, an Automation Framework
+// Copyright (C) 2026 System Initiative, Inc.
+//
+// This file is part of Swamp.
+//
+// Swamp is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License version 3
+// as published by the Free Software Foundation, with the Swamp
+// Extension and Definition Exception (found in the "COPYING-EXCEPTION"
+// file).
+//
+// Swamp is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with Swamp.  If not, see <https://www.gnu.org/licenses/>.
+
+/**
+ * In-repo regression guard for swamp-club#247: built-in command/shell must
+ * honor `--timeout` end-to-end. Drives the CLI binary against a long-running
+ * shell command and asserts the run aborts well before the natural sleep
+ * duration. Catches regressions in PRs before swamp-uat picks them up on
+ * a separate release cadence.
+ */
+
+import { assertEquals } from "@std/assert";
+import { initializeTestRepo, runCliCommand } from "./test_helpers.ts";
+import { YamlDefinitionRepository } from "../src/infrastructure/persistence/yaml_definition_repository.ts";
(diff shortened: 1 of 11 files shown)


stack72:
Closing — see swamp-club#247 for diagnosis writeup. The wiring fix works in production but the test surface needed (cross-platform shell fork behavior, sanitizer cleanup) is bigger than #247 was scoped for and deserves its own design pass.

PR #1381, opened 2026-05-13
feat(telemetry): record external-datastore-configured flag per invocation

keeb:
## Summary

- Adds `externalDatastoreConfigured: boolean` to `InvocationContext` / `InvocationContextData`, populated at telemetry-service init.
- True only when the repo marker declares a datastore whose `type` is not `filesystem` — custom providers like S3. Unset markers and local `filesystem` datastores both report false.
- Logic lives in a small `isExternalDatastoreConfigured(datastore)` helper next to `buildInvocationContext`, with unit coverage for the three cases.
- Wire shape lands at `properties.invocationContext.externalDatastoreConfigured` — `http_telemetry_sender_test` locks the contract so the swamp-club consumer can query it without a server-side change.

## Test Plan

- [x] `deno fmt --check`
- [x] `deno lint`
- [x] `deno check`
- [x] Targeted: `deno run test src/domain/telemetry/... src/cli/telemetry_integration_test.ts src/infrastructure/persistence/json_telemetry_repository_test.ts src/infrastructure/telemetry/http_telemetry_sender_test.ts` — 102 passed

### DIFF
--- a/src/cli/mod.ts
+++ b/src/cli/mod.ts
@@ -105,6 +105,7 @@ import {
   buildInvocationContext,
   clearActiveTelemetryService,
   extractCommandInfo,
+  isExternalDatastoreConfigured,
   isTelemetryDisabled,
   projectEnvSnapshot,
   setActiveTelemetryService,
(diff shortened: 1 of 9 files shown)


stack72:

PR #1383, opened 2026-05-13
feat(extensions): W7 — unify extension failure surfaces (swamp-club#342)

stack72:
## Summary

Removes the legacy `registries.<kind>.failures[]` dual failure surface and makes `aggregateState.sourceDetails[]` the single authoritative source of extension failure state in `swamp doctor extensions`. Closes the W1-W6 rearchitecture story — W7 removes the last vestige of the old failure-reporting model.

### What changed

- **Shared failure recorder** (`src/domain/extensions/source_failure_recorder.ts`) — extracted from reconcile's catch block. Both reconcile and buildIndex call the same helper, eliminating the consistency gap where buildIndex bypassed the aggregate.
- **buildIndex failure routing** — warm-start catch routes failures through `recordSourceFailure` → `repository.saveAll()`. Cold-start failures also routed, with `originalError` preserving `ValidationError` type for correct state-tag discrimination.
- **`markCatalogValidationFailed` removed** — validation failures now throw `ValidationError`, caught by the shared helper. No more direct catalog upserts that bypass the aggregate.
- **`registries.failures[]` removed** — per-registry status derived from `sourceDetails[]` failure-state tags (`ValidationFailed`, `BundleBuildFailed`, `EntryPointUnreadable`).
- **`extension_load_warnings.ts` split** — kept `emitTypeExtractionFailure` for stderr regex-mismatch warnings; removed `recordLoadFailures`, `getExtensionLoadWarnings`, and the warnings array accumulation.
- **`last_error` catalog column** — schema migration adds `last_error TEXT` to `bundle_types`. Error messages persist across process invocations. Clears on recovery to Indexed.
- **SQL state-preservation fix** — `upsert` ON CONFLICT UPDATE now conditionally excludes `state` and `last_error` when callers don't provide them, preventing `populateCatalogFromDir` from overwriting reconcile's failure states with `Indexed`.

### JSON schema changes (doctor extensions --json)

1. **Removed**: `registries.<kind>.failures[]` — the field no longer exists on `DoctorRegistryResult`
2. **Added**: `sourceDetails[].lastError?: string` — populated for failure states (`ValidationFailed`, `BundleBuildFailed`, `EntryPointUnreadable`), absent for healthy states

### Migration note

Pre-migration catalog rows in failure states get empty `lastError` until the next reconcile pass. Self-heals on first run after upgrade — not a bug, just a transient state.

### Known behavior changes (follow-ups filed)

- **Type-extraction warnings** (#351): models with non-literal `type` fields (e.g. `const TYPE = "..."`) produce stderr warnings but no longer flip doctor exit code. They don't register and don't appear in `sourceDetails[]`. Previously surfaced via the removed `registries.failures[]` pipeline.
- **Streaming protocol** (#352): `kind-completed` events carry preliminary `status: "pass"` during the scan phase; final status is computed from aggregate state in the `completed` event. Internal-only (renderer is the sole consumer).

## Test plan

- [x] 5908 unit/integration tests pass (0 failures)
- [x] `deno check` / `deno lint` / `deno fmt` clean
- [x] 24-point binary verification matrix against compiled `./swamp`:
  - Healthy model → Indexed, exit 0
  - Schema-invalid model → failure state, exit 1, lastError populated
  - Missing version → ValidationFailed, exit 1
  - Recovery: broken → fixed → Indexed, lastError clears
  - Rebundle loop guard: stable broken source converges (3 runs)
  - Mixed healthy + broken: healthy registers, broken surfaces
  - JSON contract: no `failures` field, `lastError` present
  - Vault failure surfaces in sourceDetails with correct registry status
  - lastError persists across separate process invocations
  - lastError absent from sourceDetails after recovery to Indexed
- [x] SQL regression tests: upsert without state preserves both state and last_error; upsert with explicit state overwrites both
- [x] Integration test: lastError lifecycle (populates on failure, clears on recovery)


stack72:

PR #1394, opened 2026-05-18
fix(reports): persist fallback error artifact when report execute() throws (swamp-club#365)

stack72:
## Summary

- When a report's `execute()` throws, swamp now persists a built-in fallback error artifact (markdown + JSON) instead of silently discarding the output
- Workflow status and exit code are unchanged — report failures remain advisory (a broken report must not fail an infra workflow with downstream success triggers)
- The WRN log now reads "Report X failed, fell back to default error report: ..." so users understand what happened
- Error artifacts are discoverable via `swamp data get report-<name>-json` with `{ error: true, reportName, scope, message }`

Closes swamp-club#365

## Test plan

- [x] `deno check` passes
- [x] `deno lint` passes
- [x] `deno run test` — 5959 passed, 0 failed
- [x] `deno run compile` succeeds
- [x] Verified against reproduction: workflow with throwing report shows "succeeded", exit 0, and error artifact persisted at `report-<name>-json`
- [x] Verified `swamp data get --workflow <wf> report-<name>-json` returns `{ error: true, ... }`

### DIFF
--- a/.claude/skills/swamp-report/references/report-types.md
+++ b/.claude/skills/swamp-report/references/report-types.md
@@ -121,6 +121,25 @@ reportRegistry.getByScope("method"); // Array<{ name, report }>
 
 Names must be unique — `register()` throws on duplicates.
 
+## Error Handling (Throw Semantics)
+
+If `execute()` throws, the error is **advisory** — it does not fail the workflow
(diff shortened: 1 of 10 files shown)


stack72:

PR #1397, opened 2026-05-18
fix(testing): readResource JSDoc and runtime guard for API asymmetry (swamp-club#371)

### LINKED ISSUE #371: First-class AWS SSO / IAM role assumption support in model execution
## Problem

IAM users with long-lived access keys are a deprecated pattern. AWS SSO (IAM Identity Center) and role assumption are now the standard authentication approach across organisations. Today, swamp has no first-class support for this:

- Model `globalArguments` for AWS integrations only accept static credential fields (`awsAccessKeyId`, `awsSecretAccessKey`, `awsSessionToken`)
- When an SSO session expires mid-workflow, the failure surfaces deep inside a scan step with a generic credentials error rather than a clear, actionable message
- There is no pre-flight session validation hook in the workflow execution lifecycle
- Users must manually run `aws sso login` before each workflow run and hope the session does not expire during execution

## Proposed Solution

First-class SSO/role assumption support should be a core capability, not a workaround. The following improvements would address this:

**1. AWS profile support in execution context**
Model execution should natively support an `awsProfile` concept that is resolved through the standard AWS credential chain (SSO token cache, credential process, instance profile, environment). A user should be able to declare the profile name and swamp handles the rest.

**2. Pre-flight credential validation**
Before executing any AWS-dependent model method, swamp should verify the resolved credentials are valid (e.g. via `sts:GetCallerIdentity`) and fail fast with a clear, actionable error if they are not — including a hint about how to refresh (e.g. `aws sso login --sso-session <name>`).

**3. Credential process / auto-refresh support**
Where the AWS profile is configured with a `credential_process` (e.g. `aws-sso-util`, custom scripts), swamp's execution environment should invoke it transparently at runtime rather than relying on statically resolved credentials baked into the model input.

**4. Role chaining / assume-role**
Support for specifying a role ARN to assume at execution time, using the resolved base credentials, so that cross-account access patterns work cleanly without pre-staging credentials.

## Alternatives Considered

- **Vault with credential rotation**: Storing SSO-derived short-lived credentials in a swamp vault before each run. This requires a manual pre-step, reintroduces the rotation problem the user was trying to avoid, and is operationally brittle.
- **Wrapper scripts**: Shelling out to `aws sso login` as a first workflow job. `aws sso login` is inherently interactive (browser-based); it cannot run unattended.
- **Environment variables**: Exporting credentials into the shell environment before running the workflow. Works for single-account use cases but does not scale to multi-account workflows where each job needs a different identity.

## Additional Context

This came up while building a multi-account S3 audit workflow. Five parallel scan jobs each target a different AWS account via SSO profiles. The workaround (adding `awsProfile` to the model schema and passing `AWS_PROFILE` through the execution environment) works once a session is established, but the workflow has no way to ensure the session is active, validate it before starting, or recover gracefully when it expires. SSO/role assumption is the norm in any organisation that takes security posture seriously — it should be a first-class citizen in swamp.

stack72:
## Summary

- Improved JSDoc on `readResource` in `types.ts` and `storedResources` in `test_context.ts` to clarify that `instanceName` is the `name` parameter from `writeResource`, not `specName`
- Added a runtime guard in the test context's `readResource` that throws a helpful error when `version` is a string (catches the exact misuse pattern from the issue)
- Added tests: string-version-throws guard + instanceName-vs-specName demonstration

Closes swamp-club#371

## Test Plan

- [x] `deno check` passes
- [x] `deno lint` passes
- [x] `deno run test` — 5964 passed, 0 failed
- [x] `deno run compile` succeeds
- [x] Reproduced the issue in a scratch repo, verified the fix catches the reporter's exact scenarios with a helpful error

### DIFF
--- a/packages/testing/test_context.ts
+++ b/packages/testing/test_context.ts
@@ -63,8 +63,11 @@ export interface ModelTestContextOptions {
   /** Abort signal (default: never-aborted). */
   signal?: AbortSignal;
   /**
-   * Pre-seed stored resources so readResource returns them.
-   * Keys are instance names, values are the JSON data.
+   * Pre-seed stored resources so `readResource` returns them.
+   *
+   * Keys are **instance names** (the `name` parameter from `writeResource`,
+   * not the `specName`). For example, if your model writes with
+   * `writeResource("state", "main", data)`, seed with `{ "main": data }`.
    */
   storedResources?: Record<string, Record<string, unknown>>;
   /**
@@ -200,6 +203,14 @@ export function createModelTestContext(
     instanceName: string,
     _version?: number,
   ): Promise<Record<string, unknown> | null> => {
+    if (typeof _version === "string") {
+      throw new Error(
+        `readResource(instanceName, version?) received a string as version: "${_version}". ` +
+          `Unlike writeResource(specName, name, data), readResource does not take a specName — ` +
+          `the first argument is the instance name (the "name" from writeResource). ` +
+          `Use readResource("${_version}") instead of readResource("${instanceName}", "${_version}").`,
+      );
+    }
     const data = storedResources.get(instanceName) ?? null;
     return Promise.resolve(data ? structuredClone(data) : null);
   };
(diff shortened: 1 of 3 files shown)


stack72:

PR #2188, opened 2026-08-19
fix(serve): apply --max-run-duration to all runs, not just detached

stack72:
## Summary

- `merge()` and `mergeWithConcurrency()` now skip `await Promise.allSettled(tasks)` when the abort signal has fired, so the execution service can reach its `signal.aborted` cancellation check immediately instead of blocking until all running steps finish
- Updates the `--max-run-duration` help text from "a detached run" to "a run"

**Root cause:** The `ActiveRunRegistry` timer fired correctly and aborted the controller at the configured duration, but `mergeWithConcurrency` (used by the execution service to run jobs in parallel) waited for all drain tasks to settle in its `finally` block — and those drain tasks were stuck waiting for their step generators to finish. A 30s step ran to completion despite a 5s timer.

**Verified:** compiled the binary, started `swamp serve --max-run-duration 5`, ran a `sleep 30` workflow — cancelled at 5.0s.

## Related

- Closes swamp-club#1718
- Filed swamp-club/swamp-uat#350 to un-ignore the UAT test

## Test Plan

- 3 new unit tests in `merge_test.ts` verify prompt exit on abort (52ms, not 5000ms)
- All 10,514 existing tests pass
- End-to-end verified: `swamp serve --max-run-duration 5` + `swamp workflow run sleep-step --server` aborted at 5.0s

### DIFF
--- a/src/cli/commands/serve.ts
+++ b/src/cli/commands/serve.ts
@@ -1011,7 +1011,7 @@ export const serveCommand = new Command()
   )
   .option(
     "--max-run-duration <duration:string>",
-    "Maximum wall-clock time a detached run may execute before being aborted. " +
+    "Maximum wall-clock time a run may execute before being aborted. " +
       "Accepts seconds (3600), explicit units (1h, 30m). Unset by default (no limit). " +
       "(env: SWAMP_MAX_RUN_DURATION)",
   )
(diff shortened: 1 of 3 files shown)


stack72 (commented):
**Re: CLI UX Review finding** — False positive. There is no second `--max-run-duration` help text at line 680 (or anywhere else). The `daemonEnableCommand` (lines 500–638) does not define a `--max-run-duration` option, and `collectServeExtraArgs()` (lines 312–411) does not forward it. The only definition is at line 1013, which this PR already updated.

`grep -c "max-run-duration" src/cli/commands/serve.ts` returns 3 hits, all accounted for: the option definition (line 1013), the `parseTimeout` call (line 1169), and the overflow error message (line 1176).

PR #2476, opened 2026-09-16
fix(test): pin CA:TRUE in TLS test certs so they pass under LibreSSL (swamp-club#2199)

zacharyhamm:
## Summary

Fixes swamp-club#2199. Four TLS tests relied on `openssl req -x509` emitting a v3 `basicConstraints=critical,CA:TRUE` certificate, which is only the OpenSSL 3.x default. macOS ships LibreSSL as `/usr/bin/openssl`, which emits a v1 certificate with no extensions, so on a stock macOS host:

- `hasCaTrue` returned `false` and `validateEndEntityCert` produced no `ca-true` warning
- both `tls_trust` tests failed with `UnsupportedCertVersion` instead of the expected `UnknownIssuer`

That made `deno run test` exit 1 and blocked `verify-build` on those hosts.

The fixtures now request the extension explicitly with `-addext basicConstraints=critical,CA:TRUE`:

- `src/infrastructure/runtime/tls_cert_validation_test.ts`: both CA:TRUE tests (one renamed to `hasCaTrue: returns true for CA:TRUE cert`)
- `integration/tls_trust_test.ts`: the CA step in `generateCertChain()`

Test-only change; no production code touched.

## Test Plan

- Reproduced all four failures with `PATH=/usr/bin:$PATH` (LibreSSL 3.3.6)
- With the fix, both files pass 11/11 under LibreSSL 3.3.6 and Homebrew OpenSSL 3.6.3
- Verification workflow passed on 14850dbf (build 9/9, code and adversarial review `VERDICT: pass`); attestation posted


zacharyhamm:
Superseded by #2475, which merged the same fix (plus keyUsage on the test CA) one minute earlier.

